PR #30, opened 2019-01-21
Securing POST /webhook-event endpoint

jachro:
GitLab's webhooks allow configuring a token which is sent in a request header of a call to the callback url. This PR is introducing validation of the `X-Gitlab-Token` headers value sent to the `POST /webhook-event` endpoint.

Testing the change:
`sbt clean test`


pameladelgado:
When changing the webhook token on gitlab, the error shown is 500 (internal server error) and it should be 401 (unauthorized).
To fix: Line 62 of WebhookEventEndpoint.scala : cannot decrypt authToken and does not return unauthorized error.

PR #35, opened 2019-01-29
Push event to multiple commit events

### LINKED ISSUE #36: Push event has to produce commit events for all the commits from the event
At the moment when `webhook-service` receives a Push Event from GitLab it translates it to a single Commit Event which is then stored in the Event Log. This naive implementation of the `POST /webhook-event` endpoint was initially done just to have something proving the GitLab -> Jena -> UI flow makes sense and works.
This issue is to enhance that logic so it produces Commit Events for all the commits starting from the Push Event's `after` commit id. The drawback of that solutions is that it's too greedy and produces a lot of duplicates for subsequent Push Events but there will be another issue to address that problem.

jachro:
Closes #36

This change replaces the naive push event to single commit event translation with push event to one or many commit events. The new algorithm starts to produce commit events from the push event's `after` commit id and then it looks into all its ancestors and processes them similarly. The traversal ends when there are no more parent ids.

**Testing:**
* `sbt clean test` to verify all the unit tests are passing
* UAT can be done by creating few commits and performing `git push` preferably for single commit push and multiple commits push. While push events are generated by GitLab (assuming there's a correctly created webhook for the repository) it's possible to observe  commit events generation process by:
** looking at `webhook-service`'s logs (there should be an `info` log entry for every single found commit like: `PushEvent after: <pushEvent.after>, project: <pushEvent.project.id>, CommitEvent id: <id>: stored in event log`
** `Jena`'s `renku` datastore.


rokroskar:
I tested this and it works as advertised. However, is this really recursively processing *all* of the parent commits on each commit? My test seems to indicate that this is the case... isn't this going to get very redundant very quickly?

rokroskar:
Have you looked at what Jena does with duplicate data?

jachro:
When I was checking that it seemed Jena was not creating duplicates. Whether it was updating the existing records (that's what I suspect) or simply ignoring them, I'm not sure.

Only this morning I've got an idea how to do it properly but I'd much prefer doing that as part of #37 I'm working currently on as otherwise, I'll get into a bit unpleasant merge conflicts.

rokroskar:
yep that makes sense

PR #38, opened 2019-02-01
`POST /project/id/hooks` not to create duplicate hooks and do initial loading of all the commit events to the Event Log

### LINKED ISSUE #37: `POST /project/id/hooks` not to create duplicate hooks and do initial loading of all the commit events to the Event Log
Currently, the `POST /webhooks/projects/:id` creates a new hook with a `token` which value is an encrypted Project Id to verify if a received Push Event is for a relevant project. While this is an acceptable solution from a security point of view, the idea is to enrich information stored in the token with Personal Access Token which can be used further down in the process for communication with GitLab.

As part of the issue following changes are going to be made to the endpoint:
- [x] ~~Create a Personal Access Token with the `API` scope for a project owner. The name of the token should follow a template: `<project-path>-hook-access-token`.~~
- [x] Assure the endpoint responds with Internal Server Error 500 if a hook already exists.
- [x] Create a hook token comprised of a Project Id and encrypt it.
- [x] Create a hook for a project with the encrypted token. Assure the endpoint respond with Created 201 if all goes fine.
- [x] Load Commit Events for all the commits done for the project.
- [x] Assure the endpoint responds with `Conflict 409` if project hook already exists.
- [x] Assure the endpoint responds with Internal Server Error 500 if any other problem occurs.
- [x] Assure the endpoint responds with Unauthorized 401 if either a relevant authorization header is not passed or 401 is returned from any of the calls to GitLab.
- [x] ~~Add encrypted Hook Personal Access Token to the Commit Event.~~
- [x] Make sure the Commit Events loading stops on Push Event's `commitFrom` commit id.
- [x] Fix naming of the `webhook-service`'s endpoints naming.

jachro:
Closes #37 
I think this time I did a bit better and #37 is a quite comprehensive read so I won't repeat myself here.

From a testing point of view, it's all about hitting the `POST /webhooks/projects/:projectId` endpoint. If all goes fine there should be 201 returned and apart from some log entries related to the hook creation itself, there should be also some INFO entries about all the Commit Events being pushed to the Event Log.
If the hook is already there, 409 should be returned and some ERROR statements in the log should appear. For any other errors, a 500 should be sent back to the caller and relevant ERRORs should be observed into the logs.

As part of this issue, the problem of re-sending the whole repo history to the Event Log has been addressed too. So when a Push Event is sent to the `POST /webhooks/events`, only Commit Events for commits related to that Push Event should be generated. To test it, a push to some repo having a working webhook should be done, and then log statements only for these relevant commits should appear.


jachro:
> doesn't `409` imply that something is wrong i.e. the requested webhook doesn't exist? There must be another code to signal "this wasn't actually done, but don't worry the hook is there already". Maybe just return `200` if it's there and `201` if it's newly created?

Yes, I think you are right. 409 is a bit misleading here and probably makes just confusion. When I think about that again, I'm more towards not distinguishing whether a hook was newly created or created tow months ago. I think GitLab doesn't return different codes too. But 200 and 201 makes sense. I'll change that.

jachro:
> > As part of this issue, the problem of re-sending the whole repo history to the Event Log has been addressed too. So when a Push Event is sent to the `POST /webhooks/events`, only Commit Events for commits related to that Push Event should be generated. To test it, a push to some repo having a working webhook should be done, and then log statements only for these relevant commits should appear.
> 
> Is there a reason why this is a part of the same PR? It should be separated IMO.

There's actually one reason. I wanted to have that merged that as soon as possible so Sekhar or Andreas could start their work on a final version. Apart from that, that should definitely be a separate PR.

rokroskar:
Ok, it would be easier for me to parse if they were separate, but let's leave it like this for now though it violates our own contribution guidelines.

jachro:
I know, it shouldn't be done like that and I don't like it. As I said it's just because I believe that change should go to master as quickly as possible and I was a bit afraid a new PR could get into a waiting PRs queue for a bit.

jachro:
I've just checked the OK/CREATED responses and the new endpoints on my `renku-kuba` and all seems to be fine.

webhook-service/README.md
```diff
@@ -7,10 +7,11 @@ This is a microservice which:
...
 ## API
 
-| Method | Path                               | Description                           |
-|--------|------------------------------------|---------------------------------------|
-|  GET   | ```/ping```                        | To check if service is healthy        |
-|  POST  | ```/webhook-event```               | Consumes push events sent from GitLab |
+| Method | Path                               | Description                               |
+|--------|------------------------------------|-------------------------------------------|
+|  GET   | ```/ping```                        | To check if service is healthy            |
+|  POST  | ```/webhooks/projects/:id```       | Creates a webhook for a project in GitLab |
```
rokroskar: Should this be the other way around? `projects/id/webhook`?

jachro:
Yes, I think that's better. Changed and pushed.

PR #43, opened 2019-02-07
POST /projects/:id/webhooks/validation to work for public projects

jachro:
As a `webhook-service` user I need a POST /projects/:id/webhooks/validation which checks if there's a webhook created for a project with the given id.

Testing
The endpoint should respond with:
200 - if project has the webhook
404 - if there's no webhook
401 - if the given access tokens are missing or invalid
500 - for non-public projects or if there's an error during the check

### DIFF
--- a/graph-commons/src/test/scala/ch/datascience/controllers/ErrorMessageSpec.scala
+++ b/graph-commons/src/test/scala/ch/datascience/controllers/ErrorMessageSpec.scala
@@ -20,7 +20,7 @@ package ch.datascience.controllers
 
 import org.scalatest.Matchers._
 import org.scalatest.WordSpec
-import play.api.libs.json.{JsError, JsPath, Json, KeyPathNode}
+import play.api.libs.json._
 
 class ErrorMessageSpec extends WordSpec {
 
--- a/webhook-service/README.md
+++ b/webhook-service/README.md
@@ -46,6 +46,9 @@ The endpoint requires an authorization token. It has to be
 
 #### POST /projects/:id/webhooks/validation
 
+**Notice**
+This endpoint is under development and works just for public projects. For non-public projects it responds with INTERNAL SERVER ERROR (500).
+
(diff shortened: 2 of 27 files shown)


rokroskar:
Why is this a POST? It could be just a GET on /projects/:id/webhooks

jachro:
Yeah, the problem is that this endpoint does some additional stuff which is not simply checking the webhook existence. As the name suggests it performs some validation of the access tokens and eventually it can even remove either the hook or the token or both if they are not valid. Making that short, it's in any matter idempotent.

rokroskar:
but those actions can all be done on the `/projects/:id/webhooks/` directly using different verbs (`DELETE`, `PUT`). The validation of the access token is fine, a `GET` is allowed to do that afaik.

jachro:
I mean, we could expose endpoints with different verbs but the problem is that we'll pull out the logic (which is not trivial) from the `webhook-service` to the clients. It's not the end of the world but as I need exactly the same logic in other places (e.g. `\webhooks\events`):
a) we would have to implement it at least in two services (UI/Gateway and webhook-service)
b) I'd have to expose more internals of the webhook-service for UI/Gateway or it will have to do the calls to the GitLab by itself in order to have the same logic everywhere.

In terms of the access token validation. It's not the validation of the access token you pass to request's header, it's the token which we will be storing in postgres. So effectively, the validation endpoint will be accessing the storage and eventually removing a token stored in there if it's not working anymore.

rokroskar:
hmm I see. Ok, so maybe you need the `/validation` with `POST`, but shouldn't there be a "simpler" `GET` on the `../webhooks` that just returns the active webhooks? Maybe that's not needed anywhere.

jachro:
I can always add a simple `GET /projects/:id/webhooks` when needed.

rokroskar:
I just thought that would be enough for the UI to determine what to display (cc @ableuler) But maybe you need to run the full validation anyway.

jachro:
According to what we discussed, info from a simple `GET /projects/:id/webhooks` wouldn't give UI enough information to take a decision whether to ask a user to reindex or not.

rokroskar:
got it, ok then disregard what I said :D

jachro:
No problem. A constructive discussion is always welcomed :)

jachro:
If all is fine could I ask for approval, please :)

ableuler:
Give me a moment - as I'm the one going to use these endpoints, I'll also have a look at it :)

PR #44, opened 2019-02-14
token-repository service

jachro:
As agreed with Andreas and Jiri it looks users' GitLab tokens have to be stored somewhere on the Renku Graph. As both `webhook-service` and `triples-generator` are interested in using the tokens but it doesn't look like their responsibility is to keep them, I created a simple new CRUD service for that purpose.

`token-repository` has the following API (more in the README):

| Method  | Path                               | Description                                  |
|---------|------------------------------------|----------------------------------------------|
|  GET    | ```/ping```                        | To check if service is healthy               |
|  GET    | ```/projects/:id/tokens```         | Fetches an access token for the project id   |
|  PUT    | ```/projects/:id/tokens```         | Associates the given token and project id    |
|  DELETE | ```/projects/:id/tokens```         | Deletes the token and project id association |

All of the endpoints are to be used only by internal Renku services.

It's backed up by a `postgres` database and the stored tokens are encrypted.

There is also a corresponding PR on the `renku` repository:
https://github.com/SwissDataScienceCenter/renku/pull/511


rokroskar:
Can't this just be redis? Why do we need a custom service?

ableuler:
Can you remind me of the separation of concerns between the `webhook-service` and the `triples-generator`? Would it be possible to set up things in a way that only one of the services needs access to the stored token? In this case it would be easier to avoid the extra service and just store the tokens in redis or postgres. 3217 lines of scala code for storing a token seems a bit excessive to me...

jachro:
So :) Both services `webhook-service` and `triples-generator` are completely independent, they don't know about each other anything. Both are single purpose microservices doing their own job (one dealing with the hooks, the other dealing with Event Log events). At the same time, they need the Access Tokens as they know project ids and they want to use GitLab API. Because of that effectively any of these seem to be a good fit to master the tokens data as they don't care much how that is stored, whether it's encrypted or not. So from the separation of concerns point of view, sticking the token repository logic into one of the two existing services seems not to be a good idea.

In terms of the storage, I'm guessing `postgres` is the only db which state we don't lose with a restart. 

The number of lines :) Almost 2/3 of them are tests and in fact, there are only two classes (~50 lines each) which are not adding functionality but are main classes to start the service. So if the code goes to one of the other services the savings would be probably tiny (the db access, encryption, REST endpoints and orchestration would be still there). And for the cost of these ~100 lines (to be fair there are also some lines of code for the helm charts), I believe we would make things more monolith-ish and with by far more blurred responsibility.

ableuler:
> So if the code goes to one of the other services the savings would be probably tiny (the db access, encryption, REST endpoints and orchestration would be still there).
My point was that by connecting directly to the DB, one could avoid the REST endpoints completely.

jachro:
Well, I'd say that's a bit of antipattern where you have many people trying to manage one purse plus you need to duplicate the encryption code. I believe having a clear cut between the responsibilities of our services can only help us.

ableuler:
>where you have many people trying to manage one purse plus 
That's why I was wondering if it was possible to make it one persons purse only :)

jachro:
That would be good but I don't see how this would be possible without merging triples-generator and webhook-service into one :)

PR #45, opened 2019-02-14
Made HOOK_TOKEN_SECRET be injected to the `webhook-service` on deploy

jachro:
This secret I forgot the other day and effectively all the environments would be using the same secret for encrypting the hook tokens send from Gitlab


jachro:
Done

PR #50, opened 2019-02-28
`webhook-service` has to work with private repositories

### LINKED ISSUE #42: POST /projects/:id/webhooks/validation for validating project's hook presence
#### POST /projects/:id/webhooks/validation

Validates the webhook for the project with the given `project id`. It succeeds (OK) if either the project is public and there's a hook for it or it's private, there's a hook for it and a Personal Access Token (PAT). If either there's no webhook or there's no PAT in case of a private project, the call results with NOT_FOUND. In case of private projects, if there's a hook created for a project but no PAT available (or the PAT doesn't work), the hook will be removed as part of the validation process.

**Request format**

The endpoint requires an authorization token. It has to be
- either `PRIVATE-TOKEN` with user's personal access token in GitLab
- or `OAUTH-TOKEN` with oauth token obtained from GitLab

**Response**

| Status                     | Description                                                                                                                                                       |
|----------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| OK (200)                   | When the hook exists for the project and the project is either public or there's a Personal Access Token available for it                                         |
| NOT_FOUND (404)            | When the hook either does not exists or there's no Personal Access Token available for it. If the hook exists but there's no PAT for it, the hook will be removed |
| UNAUTHORIZED (401)         | When there is neither `PRIVATE-TOKEN` nor `OAUTH-TOKEN` in the header or it's invalid                                                                             |
| INTERNAL SERVER ERROR (500)| When there are problems with validating the hook presence                                                                                                         |

jachro:
This PR makes the `webhook-service`'s endpoints to work with private repositories. The changes to the endpoints are:
- `POST /projects/:id/webhooks/validation` validates project's hook. The behaviour of the endpoint is the following:

Project visibility | Scenario | Response code | Notes
-|-|-|-
Public | Hook exists | OK (200) |
Public | Hook does not exist | NOT_FOUND (404) |
Private | The given Access Token is valid and hook exists | OK (200) | The given Access Token replaces the stored one (if there was one in the Token Repository already)
Private| The given Access Token is valid but hook does not exist | NOT_FOUND (404) | Access Token is removed from the Token Repository (if any was stored)
Private | The given Access Token is invalid but there's a valid stored Access Token and hook exists | OK (200) |
Private | The given Access Token is invalid but there's a valid stored Access Token and hook does not exist | NOT_FOUND (404) | Access Token is removed from the Token Repository
Internal |  | INTERNAL_SERVER_ERROR (500) |
Any | If anything goes wrong | INTERNAL_SERVER_ERROR (500) |

- `POST /projects/:id/webhooks` validates project's hook prior to a new hook is created using the same rules as in the grid above. Then hook is created only for cases marked with NOT_FOUND and the hook creation endpoint responds with CREATED (201). For cases marked in the grid with OK, the hook creation endpoint responds with OK, too.

- `POST /webhooks/events` - the endpoint works as before except for private repos events when the endpoint tries to use a stored Access Token. In fact, an attempt to fetch an Access Token is performed regardless of project visibility and if there's no token found, there's an assumption that the event is for a `Public` project.

closes #42


rokroskar:
@jachro is this ready to be tested?

jachro:
@rokroskar not yet. I found a bug yesterday and some noise in the logs. I addressed most of that yesterday but need to re-test today.

PR #51, opened 2019-03-05
Changing default dataset type from `mem` to `tdb`

jachro:
This is a one-liner changing default dataset type of our Renku triples store from `mem` to `tdb` so the store gets persisted to disk instead of being just in-memory.


jachro:
That was quick :)

rokroskar:
⚡️

PR #52, opened 2019-03-05
Reducing memory greediness of the services

jachro:
This PR reduces _Graph Services_ greediness of memory. I maxed the memory to 100MB per service (heap memory available for jvm process) and it looks that's enough. However, there are still three _Kubernetes_ containers created and because of that, the total number of memory used is around 1GB now.


helm-chart/renku-graph/templates/deployment.yaml
```diff
@@ -50,6 +50,8 @@ spec:
               {{- else }}
               value: {{ printf "http://$(SELF_IP):9001" | quote }}
               {{- end }}
+            - name: JAVA_OPTS
+              value: -Xmx100m
```
rokroskar: This is the initial JVM memory allocation, correct? it can get bigger?

jachro: No, that's the max. Considering these are stateless services I think it should be enough at least for now. We can always raise it up.

But please don't merge the PR as I'm still playing with that.

jachro: I'd really like to reduce memory usage of the containers too.

rokroskar: But it's only the heap so the actual consumption can be quite a bit more. Did you check the grafana values to see if that's a reasonable setting?

jachro: Yes, I was looking at various things. And yes, Xmx sets the heap size but I checked yesterday and I can happily run single Graph Microservice locally with 200M memory limit. I set the limits in helm charts at 250M for `token-repository` and `webhook-service` and 300M for `triples-generator` thinking that the last one needs some more memory for `renku log` calls. However, these values are very much guessed and we simply have to stress the services and watch what happens.

rokroskar: 👍 this is already a big improvement over the previous state!

PR #55, opened 2019-03-08
Assured all the HTTP responses from the services are JSON

### LINKED ISSUE #49: Always return JSON response
The webhook service should always return a JSON object, also on 404, for example when validating a webhook. In this case the error object should at least have a message property: https://renku.readthedocs.io/en/latest/developer/web.html#errors

jachro:
closes #49  and is described there in details.

### DIFF
--- a/graph-commons/src/test/scala/ch/datascience/http/server/HttpServerSpec.scala
+++ b/graph-commons/src/test/scala/ch/datascience/http/server/HttpServerSpec.scala
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2019 Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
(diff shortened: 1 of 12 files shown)


graph-commons/src/test/scala/ch/datascience/http/server/HttpServerSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+      .unsafeRunSync()
+
+  private implicit val timer:        Timer[IO]        = IO.timer(global)
+  private implicit val contextShift: ContextShift[IO] = IO.contextShift(global)
+
+  private val baseUri = Uri.uri("http://localhost:9999")
+  private val routes = HttpRoutes.of[IO] {
+    case GET -> Root / "resource" => Ok("response")
+  }
+  new HttpServer[IO](9999, routes).run.unsafeRunAsyncAndForget()
```
jsam:
```suggestion
  new HttpServer[IO](baseUri.port.get, routes).run.unsafeRunAsyncAndForget()
```

PR #57, opened 2019-03-11
A fix for sentry dns host not being defined

jachro:


helm-chart/renku-graph/templates/deployment.yaml
```diff
@@ -68,117 +68,117 @@ spec:
...
+      - name: JENA_BASE_URL
+        value: "http://{{ template "jena.fullname" . }}:{{ .Values.jena.service.port }}"
+      - name: JENA_ADMIN_PASSWORD
+        valueFrom:
+          secretKeyRef:
+            name: {{ template "jena.fullname" . }}
+            key: jena-admin-password
+      - name: SENTRY_DSN
+        value: {{ .Values.sentry.sentryDsnRenkuPython }}
+      - name: GRAPH_SENTRY_DNS
```
rokroskar: it should be `DSN` and not `DNS` no?

rokroskar:
@jachro can you double check the `DNS` vs `DSN` naming?

PR #59, opened 2019-03-19
File Event Log replaced with Event Log in Postgres

jachro:
This PR replaces the initial naive solution of using a shared file as the Event Log with Event Log in a relational DB (Postgres). As part of the PR following changes were made:
- webhook-service to push Commit Events derived from GitLab's Push Events to the new Event Log;
- triples-generator to pull new and re-triable Commit Events from the new Event Log and marking as being under processing;
- triples-generator to mark successfully processed Commit Events as done;
- triples-generator to mark Commit Events which triples failed while storing to Jena as re-triable;
- triples-generator to mark Commit Events which failed on stage different than storing in Jena as non-triable;
- removal of Kubernetes Persistence Volume (were previously the File Event Log was kept)


db-event-log/src/main/scala/ch/datascience/dbeventlog/commands/EventLogFetch.scala
```diff
@@ -0,0 +1,93 @@
...
+import ch.datascience.dbeventlog.{EventBody, EventStatus, IOTransactorProvider}
+import ch.datascience.graph.model.events.CommitId
+import doobie.free.connection.ConnectionOp
+import doobie.implicits._
+import doobie.util.fragments._
+
+import scala.language.higherKinds
+
+private object EventLogFetch {
+  val MaxProcessingTime: Duration = Duration.of(10, MINUTES)
```
jsam: Maybe we should put this to the environment variable.

db-event-log/src/main/scala/ch/datascience/dbeventlog/commands/EventLogFetch.scala
```diff
@@ -0,0 +1,93 @@
...
+  }.query[(CommitId, EventBody)].option
+  // format: on
+
+  private def `status IN`(status: EventStatus, otherStatuses: EventStatus*) =
+    in(fr"status", NonEmptyList.of(status, otherStatuses: _*))
+
+  private lazy val markAsProcessing: Option[(CommitId, EventBody)] => Free[ConnectionOp, Option[EventBody]] = {
+    case None => Free.pure[ConnectionOp, Option[EventBody]](None)
+    case Some((eventId, eventBody)) =>
+      sql"""update event_log
```
jsam: Does `sql` do sanitation?

db-event-log/src/main/scala/ch/datascience/dbeventlog/commands/EventLogFetch.scala
jachro: I was thinking about that but it's so deeply internal :), if you know what I mean, that probably will never be changed. I didn't want to hardcode that value somewhere between the lines and that's why I took it out to the constant.

db-event-log/src/main/scala/ch/datascience/dbeventlog/commands/EventLogFetch.scala
jachro: It doesn't when you simply run a query (I'm guessing due to performance). However, you could do proper, quite deep, sanitation on unit tests. I decided not to use that but simply run all the queries against `H2` in-memory db. I don't know what do you think but I'm kind of happy with this level of quality assurance.

PR #63, opened 2019-04-02
Scheduler for missed events

jachro:
Introduction of the DB Event Log opened quite a few new possibilities. One of them is a recovery for Push Events/Commit Events that were missed or encountered some issues during processing. This PR brings such a process into life with a scheduler checking periodically if the state of the Event Log is up to date with what's in GitLab. The check is done only for projects known by the `Graph Services` which means there was at least one event for them.

The PR is ready but needs to be merged after #60


jachro:
All in #63

rokroskar:
are you sure you don't mean #65?

jachro:
Sure, all in #65

PR #64, opened 2019-04-02
Better log statements

jachro:
I observed there was quite a lot of noise, a lot of irrelevant information and at the same time missing bits and pieces in the logs produced by the `Graph Services`. This PR tries to address these issues.

This PR is done but needs to be merged after #63


jachro:
All in #63

PR #69, opened 2019-04-30
Events processing progress fetching endpoint

jachro:
This PR introduces a new endpoint which fetches info about events processing progress of a project with the given id.

The endpoint is accessible on the webhook-service at:
`GET <webhook-service>/projects/:id/events/status`

Valid 200 response looks like:
`{ "done": 10, "total": 20, "progress": 50.00 }`

### DIFF
--- a/db-event-log/src/main/scala/ch/datascience/dbeventlog/commands/EventLogProcessingStatus.scala
+++ b/db-event-log/src/main/scala/ch/datascience/dbeventlog/commands/EventLogProcessingStatus.scala
@@ -22,16 +22,14 @@ import java.time.temporal.ChronoUnit._
 import java.time.{Duration, Instant}
 
 import cats.MonadError
-import cats.data.NonEmptyList
+import cats.data.OptionT
 import cats.effect.{Bracket, ContextShift, IO}
-import cats.implicits._
 import ch.datascience.db.DbTransactor
 import ch.datascience.dbeventlog.EventStatus._
-import ch.datascience.dbeventlog.{EventLogDB, EventStatus}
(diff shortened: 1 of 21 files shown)


jachro:
> At the moment I always get a `200` message even when the webhook has been removed (manually removed from gitlab, then called `webooks/validation`).

That's probably what we have to live with as:
* we don't consume delete hook events from GitLab
* we never clear Event Log from events we added - we treat as an audit

rokroskar:
I'm confused - on the current dev deployment you clearly get asked to add the webhook if it has been deleted

lorenzo-cavazzi:
@rokroskar yes, but
1. the user may ignore it
2. the user may remove the webhook from Github.

If the `/events/status` API returns a 404, we can cover a few annoying edge cases -- e.g. a user deletes the hook on gitlab, then he resumes working on the notebook and he goes back to the lineage page *without reloading*. He would get outdated lineage information without noticing it.

jachro:
So I've just changed the behaviour so when a user deletes the webhook manually, you'll get `404` even if there are some events in the Event Log from the past. I don't think this endpoint should be removing anything from the Event Log.

rokroskar:
agreed - see also https://github.com/SwissDataScienceCenter/renku-ui/issues/443#issuecomment-490866338

lorenzo-cavazzi:
At the moment I get the 404 response with body `"message": "Progress status for project '100' not found"` when I create the webhook. It would be nice if we could always get a 200 response with `progress: 100` when data have been indexed

jachro:
Done

PR #71, opened 2019-05-02
Graph Services acceptance tests

jachro:
This PR introduces some acceptance tests of the Graph Services. The main aim of these tests is to check and interact with the REST API of the Services as it's a black-box and though verify if the orchestration and internal communication between the Services work properly. As this is a public repository and part of the multi-module project, it doesn't call any external services but simply stubs them.


rokroskar:
why do you need a separate LICENSE file for the acceptance tests?

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/EventsProcessingStatusSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+  private val numberOfEvents = 10
+
+  feature("Status of events processing for a given project") {
+
+    scenario("As a user I would like to see progress of events processing for my project") {
+
+      val projectId = projectIds.generateOne
+
+      When("there are no events being processed")
+      Then("the status endpoint should return NOT_FOUND")
```
rokroskar: should it be `NOT_FOUND` or just `OK`?

rokroskar:
How can I run these? Is the idea that they run against a live service? If so, we should make them part of the integration test in the renku repo.

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/EventsProcessingStatusSpec.scala
jachro: Yes, it will be OK quite soon ;)

jachro:
Yes, the LICENSE should go. Copy/paste kind of issue.

jachro:
You can run the test with `sbt "project acceptance-tests" test` from the renku-graph root. These tests are intended to be Graph Services acceptance tests and thus they are focusing on: 
* verifying if all the services can start and be operational,
* verifying if the API works when called with a REST client,
* verifying the inter-services communication
They quite lightweight and quick and simply verifies if the services work as a whole.

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/WebhookValidationEnpointSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+import ch.datascience.graph.model.events.EventsGenerators._
+import ch.datascience.webhookservice.project.ProjectVisibility.{Private, Public}
+import org.http4s.Status._
+import org.scalatest.Matchers._
+import org.scalatest.{FeatureSpec, GivenWhenThen}
+
+class WebhookValidationEnpointSpec extends FeatureSpec with GivenWhenThen with GraphServices {
+
+  feature(
+    "The user can verify if there's a webhook for a project by doing POST webhook-service/projects/:id/webhooks/validation") {
```
rokroskar: `user` --> `client`?

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/WebhookValidationEnpointSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    scenario("There's a Graph Services webhook on a Public project in GitLab") {
+
+      val projectId   = projectIds.generateOne
+      val accessToken = accessTokens.generateOne
+
+      Given("project having info in GitLab")
```
rokroskar: "having" --> "has"?

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/WebhookCreationSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+      Given("latest Commit in GitLab exists")
+      `GET <gitlab>/api/v4/projects/:id/repository/commits returning OK with a commit`(projectId)
+
+      When("user issues POST webhook-service/projects/:id/webhooks")
+      val response = webhookServiceClient.POST(s"projects/$projectId/webhooks", Some(accessToken))
+
+      Then("he should get CREATED response back")
+      response.status shouldBe Created
+
+      And("the access token used in the POST should be added to the token repository")
```
rokroskar: regardless of whether the project is public or private?

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/WebhookCreationSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+import io.circe.literal._
+import org.http4s.Status._
+import org.scalatest.Matchers._
+import org.scalatest.{FeatureSpec, GivenWhenThen}
+
+class WebhookCreationSpec extends FeatureSpec with GivenWhenThen with GraphServices {
+
+  feature("The user can add a webhook to a project with POST webhook-service/projects/:id/webhooks") {
+
+    scenario("There's already Graph Services webhook on the project in GitLab") {
```
rokroskar: Isn't this going to be different for public and private projects?

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/WebhookCreationSpec.scala
jachro: Yes, because we need to use the token to get to some GitLab APIs like projects hooks.

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/WebhookCreationSpec.scala
jachro: No, it's the same for all visibilities.

PR #79, opened 2019-05-22
Allow tag pushes to trigger the build

jachro:


.travis.yml
```diff
@@ -2,6 +2,7 @@ dist: xenial
 branches:
   only:
     - master
+    - /^\d+\.\d+\.\d+$/
```
rokroskar: I think instead of doing this just remove this `branches` section altogether.

PR #82, opened 2019-06-04
Libs upgrade

jachro:
This PR addresses regular libs upgrade.

There are also fixes for some flaky tests.


acceptance-tests/build.sbt
```diff
@@ -20,4 +20,4 @@ name := "acceptance-tests"
 
 parallelExecution in Test := false
 
-libraryDependencies += "org.apache.jena" % "jena-fuseki-main" % "3.11.0" % Test
+libraryDependencies += "org.apache.jena" % "jena-fuseki-main" % "3.12.0" % Test
```
rokroskar: did you also change the version of the jena image that is used?

jachro: I did upgrade (and tested) the latest version of jena-fuseki image.

PR #88, opened 2019-06-24
knowledge-graph to handle GraphQL queries

### LINKED ISSUE #19: PoC with graphql
~~Try out https://www.hypergraphql.org/ with the Jena endpoint~~
Implement a graphql endpoint for graph queries

jachro:
This PR introduces a new graph-service exposing a new GraphQL endpoint. At the moment the endpoint allows querying project lineage only but is easily extendable to provide more information in the future.

All the necessary details about the url, request and response format could be found in the service's README file.

There are following PR related to this one on the other repositories:
* https://github.com/SwissDataScienceCenter/renku/pull/587
* https://github.com/SwissDataScienceCenter/sdsc-dev-docs/pull/32

---

closes #19


helm-chart/renku-graph/values.yaml
```diff
@@ -21,8 +21,8 @@ tokenRepository:
 
 triplesGenerator:
   image:
-    repository: renku/triples-generator
-    tag: 'latest'
+    repository: jachro/triples-generator
```
rokroskar: don't think you want to commit these

helm-chart/renku-graph/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for renku graph services
 name: renku-graph
-version: 0.1.0
+version: 0.1.0-1c62490
```
rokroskar: probably bump this to 0.2.0

jachro: Yes, I was thinking about releasing all the services but maybe once they get merged to master.

helm-chart/renku-graph/values.yaml
jachro: arghhh, again :)

helm-chart/renku-graph/Chart.yaml
rokroskar: no, I mean the working version of the chart should be `>0.1.0`, i.e. `0.2.0-1234` is a development version of `0.2.0`.

jachro: Yes, that's true -> updated.

jachro:
Hi @lorenzo-cavazzi 
I've just pushed a fix for the queries having variables. I was able to keep the scalar types as they were before but as a user, you can simply provide a primitive value. In other words, this kind of query works nicely now:
```
{ 
  "query": "query($projectPath: ProjectPath!) { 
    lineage(projectPath: $projectPath) {
      nodes { id label } 
      edges { source target } 
    } 
  }",
  "variables": {
    "projectPath": "rokroskar/vom_natt"
  }
}
```

lorenzo-cavazzi:
@jachro now it's working :tada: 
Tomorrow I'll do a few more tests before approving the PR.

lorenzo-cavazzi:
@jachro now it's working :tada: 
Tomorrow I'll do a few more tests before approving the PR.

PR #89, opened 2019-06-28
Add authentication to Jena

### LINKED ISSUE #73: do not use the admin user for Jena queries
The sparql queries currently done by the gateway are done using Jena admin credentials. We should instead create a service account for the gateway (or whatever other graph query service there might be). This could be done as a post-install script in the graph service pod.

jachro:
When I looked into #73 I realized that the Jena `renku` dataset endpoints are completely unprotected. This PR adds Basic Auth to the mentioned dataset endpoints and credentials required to access these endpoints are different than the admin credentials which now are used only to manage datasets in Jena.

This PR has two counterparts:
* https://github.com/SwissDataScienceCenter/renku/pull/589
* https://github.com/SwissDataScienceCenter/sdsc-dev-docs/pull/33
* https://github.com/SwissDataScienceCenter/testing.datascience.ch/pull/43

closes #73


helm-chart/renku-graph/charts/jena/templates/configmap.yaml
```diff
@@ -0,0 +1,33 @@
...
+    [main]
+    # Development
+    ssl.enabled = false
+
+    plainMatcher=org.apache.shiro.authc.credential.SimpleCredentialsMatcher
+    iniRealm.credentialsMatcher = $plainMatcher
+
+    [users]
+    admin={{ .Values.admin.password }}
+    renkuuser={{ .Values.renkuuser.password }}
```
rokroskar: or just `renku`?

helm-chart/renku-graph/charts/jena/templates/configmap.yaml
```diff
@@ -0,0 +1,33 @@
...
+kind: ConfigMap
+metadata:
+  name: jena-shiro-ini
+  labels:
+    app: {{ template "jena.name" . }}
+    chart: {{ template "jena.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  shiro.ini: |-
```
rokroskar: this might be better as a secret instead of configmap since it contains credentials?

jachro: I guess it would be better (and there is a secret to be used for microservice's configs) but Jena itself requires a `shiro.ini` file in her folder with the creds in open text.

rokroskar: you can mount a secret into a file all the same

helm-chart/renku-graph/charts/jena/templates/configmap.yaml
jachro: Yep, seems to be unnecessarily overcomplicated now. I'll change it.

helm-chart/renku-graph/charts/jena/templates/configmap.yaml
rokroskar: e.g. https://kubernetes.io/docs/concepts/configuration/secret/#use-case-pod-with-ssh-keys

jachro: Do you have an example somewhere? I just can't think of how this should be looking like :)

jachro: But wait there are two things I'm not sure about here:
1) the `shiro.ini` is generated with the values from the config and mounted directly into jena's filesystem so the configmap.yaml doesn't contain the passwords just the placeholders - what we would gain from turning it into a secret?
2) we need these passwords to be in both the `shiro.ini` and the 'normal' secrets.yaml so they can be injected into microservices' environment - how you can use the same values in both places? Can you have placeholders in a secret file?

rokroskar: > what we would gain from turning it into a secret?
true, atm there isn't much of a difference, except that in certain APIs secrets are hidden and only available to a user with sufficient privileges. However it's a best-practice to use secrets for confidential information. 

You can basically use exactly the same configuration you have here but make it a secret instead of a configmap. There are some differences in mounting, I believe you can look at how we mount the keycloak realm if you want an example.

jachro: Ok, let me try that

jachro: It worked!

rokroskar:
> This PR has two counterparts:

Those are three :) There should also be a PR for the changes to our continous-deployment values file.

PR #90, opened 2019-07-01
Lib upgrade

jachro:
This PR bumps the versions of dependencies


jachro:
Created from a stale branch

PR #94, opened 2019-07-03
refactor: jena users grouping

jachro:
This PR puts both Jena users under the `users` property in the `values.yaml`.

This PR has two counterparts:
* https://github.com/SwissDataScienceCenter/testing.datascience.ch/pull/45
* https://github.com/SwissDataScienceCenter/sdsc-dev-docs/pull/35


rokroskar:
This also needs a change to 

https://github.com/SwissDataScienceCenter/renku/blob/f6540b303486820f66de2b52f1bc10a52e69e14f/charts/renku/values.yaml#L595-L600

PR #100, opened 2019-07-17
Re-provision outdated triples only

jachro:
This PR changes the current functionality of re-provisioning outdated triples in the RDF Store. Instead of an endpoint deleting all the triples in the Store, there's a migration job starting after configured delay after triples-generator start-up. The job looks for outdated triples in the Store, delete all the outdated triples for a single project at once and marks the related events in the Event Log so they get picked-up for re-processing. Once a single project's triples are processed, outdated triples for another project are picked-up. The process finishes once all the outdated triples for all the projects are re-provisioned. The re-provisioning process does start only if there are no events in the Event Log waiting for processing.


acceptance-tests/src/test/resources/application.conf
```diff
@@ -9,7 +9,9 @@ projects-tokens {
   db-url-template = "jdbc:h2:mem:$dbName;DB_CLOSE_DELAY=-1;MODE=PostgreSQL"
 }
 
-triples-generator = "remote-generator"
+triples-generation = "remote-generator"
+
+re-provisioning-initial-delay = 5 seconds
```
pameladelgado: is this used anywhere?

jachro:
Yes, when the acceptance tests run, the graph services are started
programmatically and these values are picked up.

jachro:
Many thanks Pamela!

Have a great time!

PR #102, opened 2019-07-22
fix: adapted queries and tests data to use prov:agent instead of dcterms

jachro:
This is a tiny PR adapting relevant sparql queries and test data to work with prov:SoftwareAgent instead of dcterms:SoftwareAgent.

This PR needs to be merged to make triples-generator be compatible with changes from https://github.com/SwissDataScienceCenter/renku-python/issues/508.


acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/tooling/RDFStore.scala
```diff
@@ -105,8 +134,8 @@ object RDFStore {
                     |PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
                     |
                     |SELECT (COUNT(*) as ?Triples) WHERE { 
-                    |  ?agentS rdf:type <http://purl.org/dc/terms/SoftwareAgent> .
-                    |  ?agentS rdfs:label "renku $schemaVersion".
+                    |  ?agentS rdf:type <http://www.w3.org/ns/prov#SoftwareAgent> .
```
rokroskar: would it make sense to use `VALUES` here to allow for both possibilities?

jachro: Yes, could do that

jachro: Done

PR #103, opened 2019-07-23
graph-services to be auto-released on merge to master

jachro:
Graph Services have a comprehensive set of unit and acceptance tests as well as every merge to master is reviewed. Therefore, the release process should be automatic and no manual steps should be required. This PR introduces changes required to make the release like that.


.travis.yml
```diff
@@ -1,8 +1,7 @@
 dist: xenial
 branches:
   only:
-    - master
-    - /^\d+\.\d+\.\d+(-\S*)?$/
+    - /^\d+\.\d+\.\d+$/
```
rokroskar: only on tags?

jachro: yes, as effectively every successful build on master is a release you can pick up and drop to prod :)

jachro: or actually, the master still need to be there as the release process won't get kicked off :)

rokroskar: yes, that's my point :)

rokroskar: but I've been wrong about the magic of travis before as you well know... so I wasn't sure if you had some new trick :)

jachro: thankfully no tricks :)

PR #125, opened 2019-09-03
Orphan mailto:None triples to be removed on re-provisioning

### LINKED ISSUE #124: Delete orphan 'mailto:None' triples from the RDF Store
As a Renku maintainer, I'd like all the orphan 'mailto:None' triples to be deleted after the re-provisioning process is finished.

The reason why there could be some orphan 'mailto:None' triples in the RDF Store is that previous version of `renku python` was defaulting to 'mailto:None' when no user email was known. That has been fixed and a proper url resource is now used instead the 'mailto'.

jachro:
More details can be found here #124 

Closes #124


triples-generator/src/main/scala/ch/datascience/triplesgenerator/reprovisioning/OrphanMailtoNoneRemover.scala
```diff
@@ -0,0 +1,55 @@
...
+    logger:                  Logger[IO]
+)(implicit executionContext: ExecutionContext, contextShift: ContextShift[IO], timer: Timer[IO])
+    extends IORdfStoreClient[RdfDelete](rdfStoreConfig, logger)
+    with OrphanMailtoNoneRemover[IO] {
+
+  override def removeOrphanMailtoNoneTriples(): IO[Unit] = queryWitNoResult {
+    s"""
+       |PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+       |PREFIX prov: <http://www.w3.org/ns/prov#>
+       |PREFIX dcterms: <http://purl.org/dc/terms/>
```
rokroskar: the prefixes are not really being used - why not remove them?

graph-commons/src/test/scala/ch/datascience/rdfstore/InMemoryRdfStore.scala
```diff
@@ -181,4 +181,12 @@ trait InMemoryRdfStore extends BeforeAndAfterAll with BeforeAndAfter {
   }
 
   protected def runQuery(query: String): IO[List[Map[String, String]]] = queryRunner.run(query)
+
+  protected def rdfStoreSize: Int =
```
rokroskar: there is no jena API call for this?

jachro: Maybe there is but I found it easier to do a simple query :)

triples-generator/src/main/scala/ch/datascience/triplesgenerator/reprovisioning/OrphanMailtoNoneRemover.scala
jachro: Sure, I've just removed them.

graph-commons/src/test/scala/ch/datascience/rdfstore/InMemoryRdfStore.scala
rokroskar: I just wonder if this gets expensive eventually

jachro: That's just for the tests so I think it should be fine.

PR #126, opened 2019-09-04
Config errors deferred instead of being flagged up on start-up

jachro:
This PR addresses some issues about the way graph-services read values from the config files. It used to be that some config values were not read initially on service start but later on runtime. As a result, the services were starting successfully but did fail later once certain config value was read. That's clearly against the 'fail fast' rule and got fixed with this PR. A bit as a side effect, these changes resulted in quite some code simplification.


knowledge-graph/src/test/scala/ch/datascience/knowledgegraph/graphql/lineage/modelSpec.scala
```diff
@@ -38,7 +38,8 @@ class NodeSpec extends WordSpec with ScalaCheckPropertyChecks {
     }
 
     "return false if compared to a Node with a different id" in {
-      forAll { (nodeId1: NodeId, nodeId2: NodeId, nodeLabel: NodeLabel) =>
+      forAll { (nodeId1: NodeId, nodeLabel: NodeLabel) =>
+        val nodeId2 = nodeIds generateDifferentThan nodeId1
```
pameladelgado: thanks! this got rid of the test that was failing from time to time:
```
[info]   should return false if compared to a Node with a different id *** FAILED ***
[info]   TestFailedException was thrown during property evaluation.
[info]     Message: Node(t,btjCjFd) was equal to Node(t,btjCjFd)    
[info]     Location: (modelSpec.scala:42)
```

PR #128, opened 2019-09-05
Let each graph service have its own deployment

jachro:
There was a requirement of Graph Services to have separate deployments. This PR introduces that change.

There's counterpart PR on the renku repo to fix the ingress routing: https://github.com/SwissDataScienceCenter/renku/pull/652


ableuler:
I'm not sure I understand the scope of this PR correctly. I think going from 3 containers per pod exposed through 3 services to 1 deployment per service is a good idea. 
However, it seems to me that this PR is going a very long way towards deploying the 3 services through three individual charts which could also be deployed separately (which is also what the name of the PR indicates). Is this really the intention here?

jachro:
The intention of this PR is to have 3 services and 3 separate deployments for each of them so the containers are not grouped in one deployment. Whether it's achieved in the best way with my basic knowledge of charts, I'm not sure. Any suggestions/best practices more than welcomed.

ableuler:
Ok, that's how I interpreted it in the first place, but then I got unsure :)

ableuler:
I think what you want can be achieved with a much smaller amount of change. Basically, you only need to turn the containers in the old `deployment.yaml` into their own deployments. You can have a look at the gateway helm charts. There we also have two deployments (`renku-gateway` and `renku-gateway-auth`) each exposed through its own service. This should also make the changes to the other repos unnecessary because the service names which are relevant for the other components of renku to reach the graph services can remain unchanged.

jachro:
Right, now I see I should have talked to you before I started doing changes. However, once I've already done the changes and there was already `knowledge-graph` having its own deployment file and considering there are four containers with a bit long list of args, would it be fine to leave the state as it's now on the PR or you reckon it's still better to collapse them?

ableuler:
I don't mind splitting the templates for the resources over multiple files. Especially for the deployments it even seems like a good idea to have multiple files as having all three deployments in one would produce a rather long file.
However, what I find misleading are the changes to `_helper.tpl`, for example this:
```
{{- define "webhookService.chart" -}}
{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}	{{- printf "%s-webhook-service" .Chart.Name | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
{{- end -}}	{{- end -}}
```
It's still only one chart, so the the above doesn't really make sense to me.

jachro:
As I said, I'm a quite newbie to the charts and initially, I simply copied the solution from the Jena chart which I think Johann created initially. So, please could you show me how to do it in a proper way?

ableuler:
Hmm, I'm also not at the level of helm-mastery where I can give you clear instructions on how exactly to proceed without messing around a bit myself... ;) I'll give it a shot.

jachro:
Many thanks @ableuler!

PR #150, opened 2019-10-15
charts: Fix liveness probe for jena

ableuler:
An unauthenticated request to `/` on jena returns a `401` which for example on minikube (but strangely enough not on our dev cluster, maybe a k8s version question) interpreted as a failed liveness- or readiness probe. This PR fixes this by using an authentication-free endpoint.


ableuler:
> Cool! Thanks Andreas for fixing that!

My pleasure :). Please go ahead and merge it, I'm not a maintainer on this repo.

PR #151, opened 2019-10-17
re-provisioning process fails for project with old format of the project resource

jachro:
When the new format of project resources got introduced, the re-provisioning process got updated the way it can work with the new format of project resources only. This fix changes that so the process can work with all formats of project resources


jachro:
Thanks @pameladelgado!

PR #164, opened 2019-10-29
travis to publish helm charts on new snapshot creation

### LINKED ISSUE #152: Travis scripts does not trigger chartpress on new snapshot
It looks there's a bug in the renku-graph release process and chartpress does not get triggered on a new SNAPSHOT version push. In fact, chartpress should probably get triggered only on a new SNAPSHOT push but not the push preceding the tagging. Of course, chartpress should go on a new tag push.

jachro:
This PR fixes the #152 bug.

closes #152

### DIFF
--- a/.travis/travis-functions.sh
+++ b/.travis/travis-functions.sh
@@ -40,7 +40,7 @@ function createRelease() {
   git branch -u origin/${TRAVIS_BRANCH}
   git config branch.${TRAVIS_BRANCH}.remote origin
   git config branch.${TRAVIS_BRANCH}.merge refs/heads/${TRAVIS_BRANCH}
-  git config --global user.name "RenkuGraphBot"
+  git config --global user.name "RenkuBot"
   git config --global user.email "jakub.chrobasik@epfl.ch"
   git config credential.helper "store --file=.git/credentials"
   echo "https://${GITHUB_TOKEN}:@github.com" >.git/credentials


.travis/travis-release.sh
```diff
@@ -28,9 +28,9 @@ if [[ -z $TRAVIS_TAG ]] && [[ ! $TRAVIS_COMMIT_MESSAGE =~ $COMMIT_MESSAGE_PATTER
   git config branch.${TRAVIS_BRANCH}.remote origin
   git config branch.${TRAVIS_BRANCH}.merge refs/heads/${TRAVIS_BRANCH}
   git config --global user.name "RenkuGraphBot"
-  git config --global user.email  "jakub.chrobasik@eplf.ch"
+  git config --global user.email "jakub.chrobasik@eplf.ch"
```
rokroskar: Wrong email

.travis/travis-release.sh
```diff
@@ -28,9 +28,9 @@ if [[ -z $TRAVIS_TAG ]] && [[ ! $TRAVIS_COMMIT_MESSAGE =~ $COMMIT_MESSAGE_PATTER
...
   git config credential.helper "store --file=.git/credentials"
-  echo "https://${GITHUB_TOKEN}:@github.com" > .git/credentials
+  echo "https://${GITHUB_TOKEN}:@github.com" >.git/credentials
 
   # releasing graph-services
   sbt "release skip-tests default-tag-exists-answer k  with-defaults"
```
ableuler: What does this do and why is it not necessary for the cases matching the `"Setting version to .*"` pattern?

.travis/travis-release.sh
```diff
@@ -44,6 +44,22 @@ if [[ -z $TRAVIS_TAG ]] && [[ ! $TRAVIS_COMMIT_MESSAGE =~ $COMMIT_MESSAGE_PATTER
...
   ssh-add deploy_rsa
 
+  # build charts/images and push
+  cd helm-chart
+  chartpress --push --publish-chart
+  git diff
+  # push also images tagged with "latest"
+  chartpress --tag latest --push
+elif [[ -z $TRAVIS_TAG ]] && [[ $TRAVIS_COMMIT_MESSAGE =~ $COMMIT_MESSAGE_PATTERN ]]; then
+  # push the charts and images
```
ableuler: If I see this correctly, the entire code below is just copy-pasted from above. Please avoid the duplication as it will make it much easier to see what the difference in the handling of the two cases is.

.travis/travis-release.sh
jachro: fixed

.travis/travis-release.sh
jachro: No cutting corners! I thought no one will catch that :)

.travis/travis-release.sh
jachro: So as you probably know there's an auto-release process in renku-graph. It has three steps:
1. When some PR is merged, Travis builds the master's HEAD and then creates a new release using the values from `version.sbt` and `Chart.yaml`. So if it's e.g. `0.2.0-SNAPSHOT` it releases `0.2.0`. This operation comprises of:
* changing values in these two files by setting `0.2.0`, tagging the commit and pushing all of that to the master;
* changing these two files again by setting `0.3.0-SNAPSHOT` (`version.sbt`) and `0.3.0-<commit-id>` (`Chart.yaml`) and pushing again. A commit message at that point is `Setting version to 0.3.0-SNAPSHOT`.
2. Travis build get triggered for the `0.2.0` tag; this build runs chartpress and publishes `0.2.0`.
3. Another Travis build is triggered for the commit pushing `0.3.0-SNAPSHOT` to master and that publishes new helm charts for the master's HEAD. That wasn't happening before and concourse was complaining it cannot find relevant charts in our charts repo.

The check if commit message matches the `Setting version to .*` pattern is needed to distinguish if the release process should be triggered or it has just happened and so only helm charts for the master's HEAD should be built.

I'm hoping my explanation is clear enough :) If not, please shout.

.travis/travis-release.sh
fgeorgatos: `eplf.ch` -> `epfl.ch`!

.travis/travis-release.sh
jachro: All the commonalities moved to a separate script file.

.travis/travis-release.sh
jachro: Fixed in the latest push.

PR #181, opened 2019-11-08
Re-provisioning not to remove datasets linked to other projects

### LINKED ISSUE #176: triples-generator to update Person's properties on new values
As a Renku user, I'd like my name and email get updated in the KG when I change my local git settings.

Acceptance criteria:
* when new Person `schema:name` or `schema:email` get returned from the triples generation process, triples-generator should update the old values rather than inserting new ones (which causes duplicates of `schema:name` or/and `schema:email` created for that Person);

### LINKED ISSUE #180: Re-provisioning process deleting datasets linked to different projects
Unfortunately, the re-provisioning process deletes datasets which are linked to other projects. The way it should deal with the datasets is:
* remove only dataset triples with `schema:isPartOf` pointing to a project via the Commit Activity's `schema:isPartOf`; of course that's in the context of a specific commit scheduled for deletion;
* run a special post-re-provisioning clean-up deletion, removing eventual orphan datasets (datasets with no schema:isPartOf left);

jachro:
closes #180 
closes #176


acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/tooling/RDFStore.scala
```diff
@@ -131,11 +130,9 @@ object RDFStore {
...
       .map { jena =>
         val queryResults = jena.connection
-          .query("""
-                   |SELECT ?s ?p ?o WHERE {
-                   |  ?s ?p ?o .
-                   |}
-                   |""".stripMargin)
+          .query("""|SELECT ?s ?p ?o 
+                    |WHERE { ?s ?p ?o }
+                    |""".stripMargin)
```
rokroskar: I know this is just for tests but maybe putting a `LIMIT` there would make the query look less scary?

graph-commons/src/test/scala/ch/datascience/generators/CommonGraphGenerators.scala
```diff
@@ -29,7 +29,7 @@ import ch.datascience.crypto.AesCrypto
 import ch.datascience.generators.Generators._
 import ch.datascience.graph.config.RenkuBaseUrl
 import ch.datascience.graph.model.SchemaVersion
-import ch.datascience.graph.model.users.{Email, Name, Username}
+import ch.datascience.graph.model.users.{Affiliation, Email, Name, Username}
```
rokroskar: If you are trying to be complete, there is also `schema:alternateName` afaik

rokroskar: not sure if it's currently used anywhere, but I think zenodo was including it in some dataset metadata and that's why we added it.

graph-commons/src/test/scala/ch/datascience/rdfstore/triples/entities/Agent.scala
```diff
@@ -28,10 +28,10 @@ object Agent {
   {
     "@id": $id,
     "@type": [
-      "prov:SoftwareAgent",
+      "http://www.w3.org/ns/prov#SoftwareAgent",
```
rokroskar: why expand `prov` everywhere? It looks like it's in the query header anyway?

knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/ProjectDatasetsFinder.scala
```diff
@@ -48,19 +48,19 @@ private class IOProjectDatasetsFinder(
...
-       |}""".stripMargin
+    s"""|PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+        |PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+        |PREFIX schema: <http://schema.org/>
+        |PREFIX dcterms: <http://purl.org/dc/terms/>
+        |
+        |SELECT DISTINCT ?identifier ?name
+        |WHERE {
+        |  ?dataset dcterms:isPartOf|schema:isPartOf ${FullProjectPath(renkuBaseUrl, path).showAs[RdfResource]} .
+        |  ?dataset rdf:type <http://schema.org/Dataset> ;
```
rokroskar: here we might add foaf:Dataset - alternatively, we can run a query that renames all `foaf:Dataset` to `schema:Dataset`

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/tooling/RDFStore.scala
jachro: Fixed

graph-commons/src/test/scala/ch/datascience/rdfstore/triples/entities/Agent.scala
jachro: These entities are only used to generate triples for testing purposes. I needed to make them as close as possible to what `renku log` produces as triples generation does some tweaks to the JSON it receives and so it has to be very accurate.

knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/ProjectDatasetsFinder.scala
jachro: My thinking is that currently, the re-provisioning process drops all the triples from the Store and re-creates new ones. Thus, I believe all the dataset triples will be of `schema:dataset` type. Does it make sense?

PR #196, opened 2019-12-02
Paging

### LINKED ISSUE #154: GET /datasets?query=<phrase> to allow paging
As a Renku user, I'd like the `knowledge-graph/datasets` endpoint to allow paging the results.

Acceptance criteria:
* two new optional query parameters `per_page` and `page` are introduced for controlling the number of items on the page and the page number;
* if `per_page` and `page` parameters are given the results with the requested number of items from the given page are returned;
* if no `per_page` and `page` parameters are given, then the defaults of `per_page=20` and `page=1` are used;
* if the number of found items is less than the given `per_page`, a not full page is returned;
* if a non-existing page is requested, an empty result set is returned but with valid info in the pagination headers;
* if either `per_page` or `page` is a non-positive int value, a `BAD_REQUEST 400` is returned;
* there are several pagination headers returned in the response:

| Header | Description |
| ------- | ------------ |
| Total | The total number of items |
| Total-Pages | The total number of pages |
| Per-Page | The number of items per page |
| Page | The index of the current page (starting at 1) |
| Next-Page | The index of the next page |
| Prev-Page | The index of the previous page |

* there are link headers sent back with each
response. They have rel set to prev/next/first/last and contain the relevant
URL. These should be used instead of own generated URLs;
* if there's no previous page the `prev` page link is not present;
* if there's no next page the `next` page link is not present;
* the paging solution should be done in a way it can be easily re-used on other endpoints;
* as part of the issue the Renku documentation is updated with the above;

### LINKED ISSUE #155: GET /datasets?query=<phrase> to allow the query parameter to be optional
As a Renku API user, I'd like the `query` parameter on the `GET /knowledge-graph/datasets` to be optional so I can search for datasets without specifying any phrase.

Acceptance criteria:
* the `sort`, `limit` and `page` parameters should work for the found datasets;

jachro:
There's another PR with updates to the API guidelines: https://github.com/SwissDataScienceCenter/renku/pull/715

closes #154 
closes #155 

Image for testing: `'repository': 'jachro/knowledge-graph', 'tag': '27d6a76'`

An example request: `https://renku-kuba.dev.renku.ch/knowledge-graph/datasets?page=2&per_page=2&sort=name%3Aasc`


jachro:
@lorenzo-cavazzi that was a good catch! All my extensive testing didn't manage to flag that up!
So it's fixed now and ready to be looked at again :)

PR #198, opened 2019-12-04
Tokens in logs

jachro:
It was found out that for failures in `git clone` command, triples-generator logs users' access tokens. This PR fixes that.

### DIFF
--- a/helm-chart/renku-graph/Chart.yaml
+++ b/helm-chart/renku-graph/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for renku graph services
 name: renku-graph
-version: 0.32.0-6929c8b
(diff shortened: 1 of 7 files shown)


helm-chart/renku-graph/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for renku graph services
 name: renku-graph
-version: 0.32.0-a16a4c5
+version: 0.31.2-6929c8b
```
ableuler: The versions don't match here. I guess the version bot is going to take care of this after the PR has been merged?

jachro: Kind of :) Because that's a bug fix I just want a bugfix release for that change which would be `0.31.2` instead of `0.32.0` set by default by the bot.

PR #208, opened 2019-12-17
Buckets for events processing times metric

jachro:
This PR adds some custom buckets for the events processing times metric.


rokroskar:
This seems to be an non-controversial PR :)

jachro:
That's for sure :) I think we all love metrics!

PR #210, opened 2019-12-17
Renku version bug

### LINKED ISSUE #209: Reprovisioning process to look at commit agents only
It looks the re-provisioning process reaches for all Software Agent triples to find out info about the version of Renku used to generate triples. That's wrong as only commit Software Agents should be used.

jachro:
closes #209


rokroskar:
should we test this with the renku-python fix?

jachro:
I think we don't have to as the re-provisioning process looks at the Agent' ids only.

PR #217, opened 2020-01-03
Graph services cannot deal with nested groups

### LINKED ISSUE #205: Graph services cannot deal with nested groups
It looks some changes are required to handle cases with nested groups.

```
Caused by: io.circe.DecodingFailure$$anon$2: 'cramakri-group/cramakri-gp-subgroup/nested-group-project' is not a valid ch.datascience.graph.model.projects.ProjectPath: DownField(path_with_namespace)
```

jachro:
closes #205


graph-commons/src/test/scala/ch/datascience/generators/Generators.scala
```diff
@@ -165,7 +166,8 @@ object Generators {
                }
     port <- httpPorts
     host <- nonEmptyStrings()
-  } yield s"$protocol://$host:$port"
+    path <- relativePaths(maxSegments = 2)
```
rokroskar: I guess this is only used to test that absolute URLs cannot be paths? Otherwise `maxSegments` should probably not be `2`.

graph-commons/src/main/scala/ch/datascience/graph/model/projects.scala
```diff
@@ -28,8 +28,10 @@ object projects {
 
   class ProjectPath private (val value: String) extends AnyVal with RelativePathTinyType
   implicit object ProjectPath extends TinyTypeFactory[ProjectPath](new ProjectPath(_)) with RelativePath {
+    private val allowedFirstChar = ('a' to 'z') ++ ('A' to 'Z') ++ ('0' to '9') :+ '_'
+    private val regexValidator   = "^([a-zA-Z0-9_.-]+)(\\/([a-zA-Z0-9_.-]+))+$"
```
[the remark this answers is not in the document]
jachro: Thanks! Changed

graph-commons/src/test/scala/ch/datascience/generators/Generators.scala
jachro: This is a generator for property-based tests. I added some path for the URLs being generated here thinking that tests which depend on URLs check what happens if there's some path.

PR #233, opened 2020-01-16
JsonLD library for objects serialization

### LINKED ISSUE #187: JsonLD serialization mechanism
As a Renku owner, I'd like Scala code to have a mechanism of serializing objects to JsonLD.

Acceptance criteria:
* the library to be built on top of `circe`;
* a JsonLD type to be introduced;
* the JsonLD type to enforce `id` and `types` presence;
* an `asJsonLD` method to be available on data types requiring an instance of  special `JsonLDEncoder` implicitly;
* migrate all the places in Graph Services producing JsonLD object to use the new solution.

jachro:
closes #187


json-ld/README.md
```diff
@@ -0,0 +1,3 @@
+# json-ld
+
+This is JSON-LD library.
```
Panaetius: The description could be a bit more verbose ;)

jachro: Good point. Forgot about it :) Added some now.

PR #235, opened 2020-01-17
fix: Events under Processing not to be picked up before RenkuLogTimeout

### LINKED ISSUE #225: Triples Gen picks up processing events when there are still under processing
The query used by TG to poll for events has a condition classifying events having the `PROCESSING` status for longer than 10 min as leftovers from a process that died and effectively, they got re-processed. That's obviously wrong as we know there are events being processed for an hour.

The fix should take the config value of `renku-log-timeout` which is used to terminate stale `renku log` operations and use it for finding the mentioned leftovers.

jachro:
closes #225


db-event-log/src/main/scala/ch/datascience/dbeventlog/config/RenkuLogTimeout.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  import cats.MonadError
+  import cats.implicits._
+  import ch.datascience.config.ConfigLoader.find
+  import com.typesafe.config.{Config, ConfigFactory}
+
+  def apply[Interpretation[_]](
+      config:    Config = ConfigFactory.load
+  )(implicit ME: MonadError[Interpretation, Throwable]): Interpretation[RenkuLogTimeout] =
+    find[Interpretation, FiniteDuration]("renku-log-timeout", config) map RenkuLogTimeout.apply
```
rokroskar: should this be a value configured via the chart?

jachro: It actually is via the `RENKU_LOG_TIMEOUT` variable.

rokroskar: Ah ok great, I missed that thanks!

PR #242, opened 2020-01-28
fix: Fixes postgres template include and broken jena ConfigMaps

Panaetius:
The type `ConfigMap` does not have a property `type:`, only `Secret` does.

Additionally, the `postgresql` template belongs to the postgres chart in the `renku` chart requirements, so if `renku-graph` is deployed without `renku`, this fails (I needed this for my pulumi pr), so I replace it with a value that gets passed in.

This PR is needed for the pulumi change, but it will break the regular `renku` helm chart as `postgresqlHost` is not set in the main `values.yaml`. So this pr should only be merge for switching over from helm to pulumi and not before.


jachro (approved):
Looks ok so please merge it when you think that's the right time.

PR #253, opened 2020-02-21
lineage not to show renku updates

### LINKED ISSUE #169: triples-generator should not ignore deleted files
In #168 the flag `--diff-filter=d` was added to prevent `renku log` from choking on commits that delete files. Once https://github.com/SwissDataScienceCenter/renku-python/issues/790 is solved, we should remove this flag and enjoy the higher-quality KG.

jachro:
This PR:
* changes the lineage resource so it doesn't show the `renku update` nodes;
* bumps the version of `renku-python` in `triples-generator` to `0.9.1`;
Closes #169

### DIFF
--- a/acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/knowledgegraph/LineageQuerySpec.scala
+++ b/acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/knowledgegraph/LineageQuerySpec.scala
@@ -126,35 +126,24 @@ class LineageQuerySpec extends FeatureSpec with GivenWhenThen with GraphServices
 
   private lazy val theExpectedEdges = Right {
     Set(
-      json"""{"source": ${`sha3 zhbikes`.id},             "target": ${`sha8 renku run`.id}}""",
-      json"""{"source": ${`sha7 plot_data`.id},           "target": ${`sha9 renku run`.id}}""",
-      json"""{"source": ${`sha7 plot_data`.id},           "target": ${`sha12 step1 renku update`.id}}""",
-      json"""{"source": ${`sha7 clean_data`.id},          "target": ${`sha8 renku run`.id}}""",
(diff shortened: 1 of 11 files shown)


rokroskar:
> It also upgrades renku-python in triples-generator to 0.9.1

which doesn't exist yet... ?

jachro:
I know, I was trying to envisage the future 😄 I'll merge it when it's ready.

lorenzo-cavazzi:
I tested it and I got conflicting results.
On a "clean" environment where I build the KG from scratch, this works fine
https://lorenzotest.dev.renku.ch/projects/lorenzo.cavazzi.tech/flights-tutorial/files/lineage/data/201901_us_flights_1/2019-01-flights.csv.zip
Screenshot_20200226_151351


But on an old one where I updated the graph component, I got something dirty with a duplicate node :thinking: 
https://lorenzo.dev.renku.ch/projects/lorenzo.cavazzi.tech/flights-tutorial/files/lineage/data/201901_us_flights_1/2019-01-flights.csv.zip
Screenshot_20200226_151414

jachro:
So the reason why there are these two `renku run .. FilterFlights.py` nodes is that because of some reason there are two different CWL files for that `ProcessRun`: `.renku/workflow/81e36868d9d349908d50b11c9d7a6970_python.cwl` and `.renku/workflow/74971acd0f494f74a02422652bfcac3a_python.cwl`. Did you maybe executed `renku run python src/00-FilterFlights.py data/201901_us_flights_1/2019-01-flights.csv.zip data/output/2019-01-flights-filtered.csv` twice? If not, @rokroskar do you maybe know why or in what conditions, we generate yet another CWL file for the same command? I can try to collapse these but would like to know first if that behaviour is something normal or rather not.

rokroskar:
My guess is it was run twice.

lorenzo-cavazzi:
It was, but then why two different results on two different renku deplyoments for the same project?

jachro:
That's because `renku log` failed for some of the Commit Events.

PR #273, opened 2020-03-11
TG to call renku migrate before renku log

### LINKED ISSUE #263: triples-generator to be able to work in the presence of new renku-python migrations
As a Renku user, I'd like triples-generator to work with new migration mechanism in renku-python.

Acceptance criteria:
* triples-generator to run project migration with `renku migrate --no-commit` as a prerequisite to `renku log`;
* the migration call should be done as a separate runtime call so it can be reported properly in the logs;
* migration failures should be classified as non-recoverable errors and should stop further steps in the triples generation process.

jachro:
**Testing scenario:**
Unfortunately, to test this change some help from @jachro is essential. The reason for that is the change requires deploying a new version of triples-generator using renku-python v0.9.2. Once triples-generator is deployed, verification of the new data pushed into the KG can be done by checking the progress of re-provisioning process on the **Event Log events** graph from https://grafana.dev.renku.ch/d/cb9J1WBZz/event-log?orgId=1&refresh=30s&from=now-1h&to=now&var-kube_env_name=renku-kuba-renku-triples-generator. Important is to check if the number of `NON_RECOVERABLE_FAILURE` does not go up to much (currently, there are 3.909K events in total of which only 28 failed).

**Image for testing:**
```
'repository': 'jachro/triples-generator', 'tag': '96bdae2e'
```

closes #263


rokroskar:
I guess this needs a new renku-python release, so should it be put on hold for the time being?

jachro:
Yes, you are right. How do you put an issue on hold? Do you label it somehow?

rokroskar:
We have a "do-not-merge" label - I added it.

PR #297, opened 2020-04-21
Event log

### LINKED ISSUE #265: Memory leak probably somewhere in the Prometheus lib
It looks there's a memory leak probably somewhere in the Prometheus lib.
```
java.lang.OutOfMemoryError: Java heap space
    at java.util.Arrays.copyOf(Arrays.java:3332)
    at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:124)
    at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:448)
    at java.lang.StringBuffer.append(StringBuffer.java:270)
    at java.io.StringWriter.write(StringWriter.java:101)
    at io.prometheus.client.exporter.common.TextFormat.write004(TextFormat.java:40)
    at org.http4s.metrics.prometheus.PrometheusExportService$.$anonfun$generateResponse$1(PrometheusExportService.scala:38)
    at org.http4s.metrics.prometheus.PrometheusExportService$$$Lambda$1436/1543602907.apply
    at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:87)
    at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:355)
    at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:376)
    at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:316)
    at cats.effect.internals.IOShift$Tick.run(IOShift.scala:36)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:748)
```

After some investigation, it looks the reason for that is probably an excessive creation of new threads due to overuse of the `*>` operator. As the problem mostly occurs in triples-generator, the issue will be addressed as part of #270.

### LINKED ISSUE #268: Turn the Event Log lib into a service
As a Renku owner, I'd like an Event Log service to be the API of the Event Log queue instead of the current shared library.

Reasoning:
* Currently, there's no easy way of throttling triples generation processes per Renku deployment as each triples-generator service uses its own throttling. However, in the case of multiple triples generator instances, such a way of controlling the triples generation processes does not work correctly, as lack of a single decision point (a dispatcher) causes an excessive number of calls to the external APIs e.g. GitLab. This problem will be even of higher importance once the triples generation process is not dependent on calling the time-consuming `renku log` command but will be working of small graph diffs read directly from GitLab.
* There's no reasonable way of gathering the Event Log metrics. At the moment the metrics are calculated by polling the Event Log DB by each triples-generator instance.

### LINKED ISSUE #269: Wrap the Even Log lib into a REST service
As a Renku user, I'd like the Event Log lib to become a REST service.

Acceptance criteria:
- [x] turn the Event Log lib into a REST ųservice;
- [x] create a `Dockerfile`;
- [x] create a helm chart for the service;
- [ ] ~expose relevant functionalities as endpoints;~ will be done as part #270
- [x] make the Event Log metrics be served by the service;

### LINKED ISSUE #270: Made the renku-graph services to use the Event Log service instead of the lib
As a Renku owner, I'd like all the Event Log API interactions in all the renku-graph modules to be done through the REST API not by using the lib.

Acceptance criteria:
- [x] find usages of the Event Log lib in the renku-graph modules and replace them with the relevant calls to the event-log service REST APIs;
- [x] refactor event-log service;
- [x] refactor the `events_waiting_count` metric so it's incremented and decremented by API calls rather than backed by DB queries;
* - [x] rename `db-event-log` to `event-log`;
* - [x] think of adding relevant links to the endpoints responses;
* - [x] rename command classes and method names and put them into proper packages;
* - [x] take out project dependency to event-log from webhook-service and triples-generator;
* - [x] clean-up webhook-service and triples-generator helm charts (remove unused env variables);
* - [x] get rid of event-log DB transactors in webhook-service and triples-generator;
* - [x] check for eventual TODOs;
- [ ] ~raise an issue for Gateway to call event-log service to find processing status;~
- [ ] ~raise an issue for webhook-service to remove the processing status;~

### LINKED ISSUE #295: TG should use CLEAR DEFAULT instead of DELETE to remove all triples
It looks like wiping out the whole renku dataset with a `DELETE` query can take quite some time and resources. Triples-generator should be using `CLEAR DEFAULT` instead.

jachro:
This PR replaces usages of event-log lib with REST calls to the new event-log microservice.

**Testing scenarios:**
* effectively, a full regression needs to be done

**Images for testing:**
```
'repository': 'jachro/triples-generator', 'tag': 'ac234351'
'repository': 'jachro/event-log', 'tag': 'ac234351'
'repository': 'jachro/token-repository', 'tag': '1314deff'
'repository': 'jachro/webhook-service', 'tag': '26822021'
'repository': 'jachro/knowledge-graph', 'tag': '1314deff'
```

closes #268
closes #269
closes #270 
closes #265 
closes #295


event-log/README.md
```diff
@@ -0,0 +1,243 @@
+# event-log
+
+This is a microservice which provides CRUD operations for Event Log DB.
+
+## API
+
+| Method | Path                                     | Description                                                    |
+|--------|------------------------------------------|----------------------------------------------------------------|
+|  POST  | ```/events```                            | Creates an event with a `NEW` status                           |
+|  PATCH | ```/events```                            | Changes events' data by applying the given patch               |
```
ableuler, line 10: I think this should be `PATCH /events/:id`

event-log/README.md
```diff
@@ -0,0 +1,243 @@
...
+|  GET   | ```/events/latest```                     | Finds events for all the projects with the latest `event_date` |
```
ableuler, line 11: How about `GET /events?order_by=event_date`

event-log/README.md
```diff
@@ -0,0 +1,243 @@
...
+|  PATCH | ```/events/:id/projects/:id/status```    | Updates event status                                           |
```
ableuler, line 12: Do I understand that each event is uniquely linked to one project? If so, I think it can simply be a `PATCH` to `/events/:id/`.

event-log/README.md
```diff
@@ -0,0 +1,243 @@
...
+|  GET   | ```/events/projects/:id/status```        | Finds processing status of events belonging to a project       |
```
ableuler, line 13: This could be `GET /events?project_id=<project-id>`, parsing the response to get the processing stati should in this case be the responsibility of the client in my opinion.

event-log/README.md
```diff
@@ -0,0 +1,243 @@
...
+|  POST  | ```/events/subscriptions?status=READY``` | Adds a subscription for the events                             |
```
ableuler, line 14: I would suggest to make the subscription a top-level entity, so just `POST /subscriptions`. Details of the subscriptions can be in the body.

event-log/README.md
ableuler: Actually, [our own docs](https://renku.readthedocs.io/en/latest/developer/contributing/api-guidelines.html#sorting) suggest `GET /events?sort=event_date:ASC`

event-log/README.md
jachro: Yes, this endpoint modifies all the events.

event-log/README.md
jachro: I quite like the idea of sorting. The only thing not that obvious to me is how to make the resource to somehow explain that the list you get contains one event (with the most recent date) per project. In other words, you get a list of the `project -> latest event` tuples. Any idea?

event-log/README.md
jachro: That would be nice but unfortunately, an event is uniquely identified by a compound key `event id -> project id` as potentially there could be two events with the same `event id` belonging to two different projects.

event-log/README.md
jachro: Your suggestion made me think that there's a bigger problem with that resource. The naming suggest that it simply returns some project events' data but in fact, it gives very specific info about the processing status of events belonging to a project. Wouldn't that then be better if we go with smth like this:
`GET /events/processing-status?project_id=<project-id>`
?

event-log/README.md
jachro: Great idea!

event-log/README.md
rokroskar: that doesn't seem too restful unless you have `/events/:id/processing-status` and you get the `:id` from `/events?project_id=<project-id>`

jachro: So would maybe `/events/projects/:id/processing-status` work the best? The way I see it is that we don't filter the results here but simply return some specific data for some events of a certain project.

rokroskar: but there is also `/events/:id/projects/:id/status` - I think this mixing makes it confusing about which belongs to what

jachro: Agreed. What about `/events/projects/:id/processing-progress`?

rokroskar: I think this gets us back to @ableuler's initial question - why not just put the `processing-progress` into the body? then the endpoint can be just `/events/projects/:id`

jachro: Well, a `GET` with a body does not seem that right. Here's what people say in StackOverflow citing HTTP specs:

```
The HTTP specification says in section 4.3

A message-body MUST NOT be included in a request if the specification of the request method (section 5.1.1) does not allow sending an entity-body in requests.

Section 5.1.1 redirects us to section 9.x for the various methods. None of them explicitly prohibit the inclusion of a message body. However...

Section 5.2 says

The exact resource identified by an Internet request is determined by examining both the Request-URI and the Host header field.

and Section 9.3 says

The GET method means retrieve whatever information (in the form of an entity) is identified by the Request-URI.
```

rokroskar: sorry I didn't mean in the request body, I meant that it would be returned in the list of events and their properties.

ableuler: I think @rokroskar meant the response and not the request body. For me there are two clean RESTful options: 
- `GET /events?project_id=<project-id>` which should return a list of events, the processing status would be one field of an event.
- `GET /statuses?project_id=<project-id>` which returns a list of statuses, statuses therefore become resources just like events.

event-log/README.md
ableuler: Ah, I see. Is it then fair to say that the event Id could be considered an event type? I think the accepted way to handle composite ids in REST would be 
`/events/:event-type/:project-id`

event-log/README.md
ableuler: Ok - in this case I wouldn't be too dogmatic and leave it as you suggested.

event-log/README.md
ableuler: Then I would just go for `GET /events?latest_per_project=true`

event-log/README.md
jachro: `/processing-status?projects-id=<id>`

PR #303, opened 2020-05-04
All triples removal bug

### LINKED ISSUE #302: Query removing all the triples does not work
It looks like the new query removing all the triples during the re-provisioning process does not work and Jena throws the following error:
```
ERROR application - Re-provisioning failure
ch.datascience.http.client.RestClientError$UnexpectedResponseException: POST http://renku-jena:3030/renku/update returned 500 Unknown escape: \3; body: Error 500: Unknown escape: \3
```

jachro:
As described in #302 the `CLEAR DEFAULT` operation does not seem to work in all Jena instances. This PR fixes that with a simple `DELETE` query removing batches of triples until the dataset is empty. The batching approach is needed from Jena's health point of view as it can simply blow up when trying to remove all the triples in one go.

**Testing image:**
`'repository': 'jachro/triples-generator', 'tag': 'f2a52b79'`
This image was already successfully tested on `renku-kuba`.

**Testing scenarios:**
* simply drop mentioned image to an environment of your choice and observe the [Grafana Event Log metrics](https://grafana.dev.renku.ch/d/cb9J1WBZz/event-log?orgId=1&from=now-1h&to=now&refresh=30s&var-kube_env_name=renku-kuba-renku) to see if all the events get re-scheduled for re-processing.

closes #302

### DIFF
--- a/triples-generator/src/main/scala/ch/datascience/triplesgenerator/reprovisioning/TriplesRemover.scala
+++ b/triples-generator/src/main/scala/ch/datascience/triplesgenerator/reprovisioning/TriplesRemover.scala
@@ -29,6 +29,10 @@ private trait TriplesRemover[Interpretation[_]] {
   def removeAllTriples(): Interpretation[Unit]
 }
 
+private object TriplesRemover {
+  val TriplesRemovalBatchSize: Long = 50000
+}
+
 private class IOTriplesRemover(
     rdfStoreConfig:          RdfStoreConfig,
     logger:                  Logger[IO],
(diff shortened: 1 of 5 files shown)


pameladelgado:
is the tag to test this still `f2a52b79` ?

jachro:
Yes, that should be the one.

triples-generator/src/main/scala/ch/datascience/triplesgenerator/reprovisioning/TriplesRemover.scala
```diff
@@ -29,6 +29,10 @@ private trait TriplesRemover[Interpretation[_]] {
   def removeAllTriples(): Interpretation[Unit]
 }
 
+private object TriplesRemover {
+  val TriplesRemovalBatchSize: Long = 50000
```
pameladelgado, line 33: how did you come up with this number?

jachro: Don't ask. Not to big and too small :)

PR #329, opened 2020-06-30
Handling new JsonLD model workflow 

### LINKED ISSUE #313: Lineage endpoint to work with workflow info storing changes
As a renku-graph API user, I'd like the `POST /knowledge-graph/graphql` (lineage resource) to work in the presence of the changes done to the way workflow info is stored (check [here](https://github.com/SwissDataScienceCenter/renku-python/pull/1185) for more details).

Acceptance criteria:
* update the entities in renku-graph to fit the new modelling;
* the lineage resource served by the knowledge-graph's GraphQL endpoint (`POST /knowledge-graph/graphql`) to work as before the workflow info storing changes;
* are there more things to adjust???

joke1196:
closes #313


triples-generator/Dockerfile
```diff
@@ -23,9 +23,12 @@ ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 ENV TZ UTC
 
 # Installing Renku and other dependencies
-RUN apk add --no-cache tzdata git git-lfs curl bash python3-dev openssl-dev libffi-dev linux-headers gcc libxml2-dev libxslt-dev libc-dev && \
-    python3 -m pip install 'renku==0.10.5' sentry-sdk && \
+RUN apk add --no-cache tzdata git git-lfs curl bash python3-dev py3-pip py3-wheel openssl-dev libffi-dev linux-headers gcc libxml2-dev libxslt-dev libc-dev yaml-dev && \
+    python3 -m pip install 'git+https://github.com/SwissDataScienceCenter/renku-python.git@1298-calamus-workflows' sentry-sdk && \
     chown -R daemon:daemon . && \
+    export RENKU_DISABLE_VERSION_CHECK=true && \
+    git config --global user.name 'renku'  && \
```
rokroskar, line 30: why do the username and email need to be added to git config? you're not making any commits to the repos, are you?

jachro: Oh, actually, we do create commits. That happens when we call `renku migrate` but we don't push them back to the repo. The reason we had to do it is due to the `invalid email` error which happens if there's no valid email on a commit activity's creator. Such activity could be eventually created during migration. However, according to Ralf, this info is not used later and as far as I remember it does not ends up in the generated json-ld.

rokroskar: Thanks for clarifying!

PR #338, opened 2020-07-06
Fix several bugs in the helm templates

ableuler:
Fix bugs in templates that led to validation problems and ignored resource limits.

### DIFF
--- a/helm-chart/renku-graph/charts/jena/templates/deployment.yaml
+++ b/helm-chart/renku-graph/charts/jena/templates/deployment.yaml
@@ -58,7 +58,7 @@ spec:
               mountPath: /docker-entrypoint.sh
               subPath: docker-entrypoint.sh
           resources:
-{{ toYaml .Values.resources | indent 12 }}
+{{- toYaml .Values.resources | nindent 12 }}
       volumes:
         - name: docker-entrypoint
           configMap:
--- a/helm-chart/renku-graph/charts/jena/templates/lucene-config.yaml
+++ b/helm-chart/renku-graph/charts/jena/templates/lucene-config.yaml
@@ -8,7 +8,6 @@ metadata:
     chart: {{ template "jena.chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
-type: Opaque
 data:
   {{ .Values.global.graph.jena.dataset }}.ttl: |-
 
--- a/helm-chart/renku-graph/charts/jena/templates/docker-entrypoint.yaml
+++ b/helm-chart/renku-graph/charts/jena/templates/docker-entrypoint.yaml
@@ -8,7 +8,6 @@ metadata:
     chart: {{ template "jena.chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
-type: Opaque
 data:
   docker-entrypoint.sh: |-
     #!/bin/sh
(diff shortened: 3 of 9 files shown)


helm-chart/renku-graph/charts/jena/templates/lucene-config.yaml
```diff
@@ -8,7 +8,6 @@ metadata:
     chart: {{ template "jena.chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
-type: Opaque
 data:
```
rokroskar, line 11: I guess it should be a list?

ableuler: no, what I wanted to say is that config-maps don't have types at all.

helm-chart/renku-graph/charts/jena/templates/deployment.yaml
```diff
@@ -58,7 +58,7 @@ spec:
               mountPath: /docker-entrypoint.sh
               subPath: docker-entrypoint.sh
           resources:
-{{ toYaml .Values.resources | indent 12 }}
+{{- toYaml .Values.resources | nindent 12 }}
```
rokroskar, line 61: why put it on a new line when you changed `indent` to `nindent`? If you are leaving it as is, then there is no need to use `nindent`.

ableuler: I just used what the standard helm template for new charts suggests (ie what you get from helm `helm create`).
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ include "foo2.fullname" . }}
  labels:
{{ include "foo2.labels" . | indent 4 }}
spec:
  replicas: {{ .Values.replicaCount }}
  selector:
    matchLabels:
      app.kubernetes.io/name: {{ include "foo2.name" . }}
      app.kubernetes.io/instance: {{ .Release.Name }}
  template:
    metadata:
      labels:
        app.kubernetes.io/name: {{ include "foo2.name" . }}
        app.kubernetes.io/instance: {{ .Release.Name }}
    spec:
    {{- with .Values.imagePullSecrets }}
      imagePullSecrets:
        {{- toYaml . | nindent 8 }}
    {{- end }}
      containers:
        - name: {{ .Chart.Name }}
          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
          imagePullPolicy: {{ .Values.image.pullPolicy }}
          ports:
            - name: http
              containerPort: 80
              protocol: TCP
          livenessProbe:
            httpGet:
              path: /
              port: http
          readinessProbe:
            httpGet:
              path: /
              port: http
          resources:
            {{- toYaml .Values.resources | nindent 12 }}
      {{- with .Values.nodeSelector }}
      nodeSelector:
        {{- toYaml . | nindent 8 }}
      {{- end }}
    {{- with .Values.affinity }}
      affinity:
        {{- toYaml . | nindent 8 }}
    {{- end }}
    {{- with .Values.tolerations }}
      tolerations:
        {{- toYaml . | nindent 8 }}
    {{- end }}
```

ableuler: But I agree with you that it seems unnecessary...

rokroskar: So does this render with extra blank lines? Or does it skip the newline if it's already on a new line? I guess it doesn't really matter - it's also a bit safer in case the template expression is moved up?

ableuler: It renders just the way you'd write the yaml:
```
          resources:
            limits:
              memory: 750Mi
```
The reason is that the `{{- ` doesn't just strip whitespace but also newlines.

PR #340, opened 2020-07-07
fix: added logs when git clone commands fails

joke1196:


triples-generator/src/main/scala/ch/datascience/triplesgenerator/eventprocessing/triplesgeneration/renkulog/Commands.scala
```diff
@@ -112,7 +114,9 @@ private object Commands {
           case err if recoverableErrors exists err.contains =>
             GenerationRecoverableError(errorMessage(err)).asLeft[Unit].pure[IO]
           case err =>
-            new Exception(errorMessage(err)).raiseError[IO, Either[GenerationRecoverableError, Unit]]
+            logger.error(errorMessage(err))
```
jachro, line 117: We're you able to verify if that works better now? My impression was that there's a problem with the `errorMessage(err)` which simply truncates the first line from the exception so we end up with log entries like `got clone: `. I think of you deploy your version and try to generate triples for nonexisting project you did be able to verify the behaviour. In terms of logging the error in the clone method directly, I think we can end up with the clone error statements twice in the log file as I believe we can log the exception somewhere where the clone command is invoked but don't remember for sure.

joke1196: Yes you are right there is an issue with how we capture the result. I think we should read the std error stream instead of std out.

jachro: I believe so. Or even read from both outs and merge them.

PR #345, opened 2020-07-27
build: allow renku-python version override

### LINKED ISSUE #161: Parametrize renku-python version in renku-graph builds
As a Renku owner, I'd like the version of renku-python in the renku-graph build to be parametrized.

Acceptance criteria:
* version of renku-python to be specified in the build's env variables; see e.g. https://github.com/SwissDataScienceCenter/renku-jupyter/blob/e3e639c379acba1392ad16e93b5d2f8f68dbca27/.travis.yml#L34-L40;
* triples-generator docker image building process should use the build's env variable with the renku-python version. See e.g. https://github.com/SwissDataScienceCenter/renku-jupyter/blob/e3e639c379acba1392ad16e93b5d2f8f68dbca27/docker/base/Dockerfile#L60-L61

rokroskar:
closes #161 

The intended use here is that it shouldn't be necessary to rebuild the triples generator image to test compatibility with a development version of renku-python. Note that the "version" should be  a fully-qualified pip-installable target, e.g.

```
renku==0.10.5
git+https://github.com/swissdatasciencecenter/renku-pyhon@<branch>#egg=renku
```

This image is currently deployed as the TG of rok.dev.renku.ch.

### DIFF
--- a/helm-chart/renku-graph/values.yaml
+++ b/helm-chart/renku-graph/values.yaml
@@ -41,6 +41,8 @@ triplesGenerator:
   ## a demanded number of concurrent triples generation processes
   generationProcessesNumber: 4
   threadsNumber: 6
+  # set this to a pip-installable renku-python version which will be installed on startup
+  renkuPythonDevVersion:
 
 webhookService:
   image:
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@
 .travis*
 helm-chart
 *.enc
+**/Dockerfile
--- a/helm-chart/chartpress.yaml
+++ b/helm-chart/chartpress.yaml
@@ -1,5 +1,6 @@
 charts:
   - name: renku-graph
+    resetTag: latest
     imagePrefix: renku/
     repo:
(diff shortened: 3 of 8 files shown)


helm-chart/renku-graph/values.yaml
```diff
@@ -41,6 +41,8 @@ triplesGenerator:
   ## a demanded number of concurrent triples generation processes
   generationProcessesNumber: 4
   threadsNumber: 6
+  # set this to a pip-installable renku-python version which will be installed on startup
+  renkuPythonDevVersion:
```
jachro, line 45: Would that work if you don't specify a value in your `values.yaml`?

rokroskar: this just means it's null

rokroskar:
I still haven't gotten it to show that it's ready... let me try to redeploy

rokroskar:
Ok I tested this now it works fine - I just had to delete the deployment, it was stuck in some weird state.

PR #346, opened 2020-07-30
Immutable datasets

### LINKED ISSUE #306: KG dataset resources to work with immutable datasets
As a Renku API user, I'd like KG dataset resources to show the latest versions of datasets if they are modified.

**Acceptance criteria:**
* `GET /knowledge-graph/datasets` to look for matches of the given criteria only among the latest datasets versions; the latest version of a dataset is understood as one without other datasets having the `prov:wasDerivedFrom` pointing to it;
* `GET /knowledge-graph/datasets/:id` to return details of the dataset with the given `identifier`; it's acceptable that if the given `id` is not the `id` of the latest version of the dataset, details of that particular version will be returned;
* `GET /knowledge-graph/projects/:namespace/:name/datasets` to return all project's datasets in the versions the project uses them; that does not necessarily mean the datasets need to be in the latest version exists as for instance, certain datasets could have newer versions on the projects where they were originally created;
* for the `GET /knowledge-graph/datasets/:id` and `GET /knowledge-graph/projects/:namespace/:name/datasets`:
  * the `sameAs` property:
    * will be present if it's set explicitly on the dataset (a case when a dataset was imported and not modified in the project);
    * will be present even if it's not set explicitly on the dataset (a case when a dataset was modified in the project); its value will be a reference to this dataset's `id`;
  * additional `derivedFrom` property will be added when the dataset has some previous version(s) (it's a modified dataset); the value of the property will be a reference to the previous version of the dataset (its `id`);
  * additional `projectIdentifier` property will be added showing dataset identifier within a project (mapped from `https://swissdatasciencecenter.github.io/renku-ontology#identifier`);
    * the `projectIdentifier` value does not change for a dataset within a project and when the dataset is modified it holds its value;
    * the `projectIdentifier` value gets a new value when a dataset is imported to another project;
    * multiple datasets can have the same `projectIdentifier` value only if there was a dataset on some project and that project got forked.

**Things to consider:**
* does the insertion of additional aggregate triples make sense?
  * potentially useful aggregates: `topmostSameAs` and `topmostWasDerivedFrom`;
  * problem of several events being processed at once;
  * problem of events not being processed in order;
    * the aggregate preparation would have to update not only the current dataset details but also all the datasets which were imported from it (and all their descendants);
  * problem of updating the aggregates for other projects;

jachro:
This PR brings graph-services to work in the presence of immutable datasets.

closes #306


helm-chart/renku-graph/charts/jena/templates/lucene-config.yaml
```diff
@@ -11,43 +11,44 @@ metadata:
...
   {{ .Values.global.graph.jena.dataset }}.ttl: |-
 
-    @prefix :        <#> .
-    @prefix fuseki:  <http://jena.apache.org/fuseki#> .
-    @prefix rdf:     <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
-    @prefix rdfs:    <http://www.w3.org/2000/01/rdf-schema#> .
-    @prefix tdb:     <http://jena.hpl.hp.com/2008/tdb#> .
-    @prefix ja:      <http://jena.hpl.hp.com/2005/11/Assembler#> .
-    @prefix text:    <http://jena.apache.org/text#> .
+                                                 @prefix:        <#> .
```
rokroskar, line 14: is this indented so far on purpose?

joke1196: No this is a side effect of the scala formatter.

joke1196 (approved):

PR #350, opened 2020-08-12
Fix for a CI and docker issues

jachro:
It looks the docker version on `chartpress` is not pinned so the CI builds stopped to work when a new python docker lib got released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_install:
   - sudo apt-get update
 install:
   - travis_retry sudo pip3 install -U pip
-  - travis_retry sudo pip3 install "chartpress==0.3.2" "ruamel.yaml==0.15.54"
+  - travis_retry sudo pip3 install "docker==4.2.2" "chartpress==0.6.0" "ruamel.yaml==0.15.54"
   # Installing Helm
   - wget -q ${HELM_URL}/${HELM_TGZ}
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -36,7 +36,7 @@ before_install:
   - sudo apt-get update
 install:
   - travis_retry sudo pip3 install -U pip
-  - travis_retry sudo pip3 install "chartpress==0.3.2" "ruamel.yaml==0.15.54"
+  - travis_retry sudo pip3 install "docker==4.2.2" "chartpress==0.6.0" "ruamel.yaml==0.15.54"
```
joke1196, line 39: Is the version 4.3.0 of docker not compatible? I just saw they've released this version on the 10th of August

rokroskar: does it not work with `chartpress==0.3.2` and `docker==4.2.2`?

jachro: I guess it would work with `chartpress==0.3.2` and `docker==4.2.2` but I tried out the most recent version of `chartpress` and all seems to be fine, too.

rokroskar: I think it works here because you are always supplying a tag with `--tag`. But in the other repos we don't and chartpress changed how they generate the tags so for the other repos we have to rethink how the tag is used.

jachro: I see. I wasn't aware of that.

PR #359, opened 2020-08-19
FIX: Java Instant - SQL Date conversion - closes #354

### LINKED ISSUE #354: Event-log fails when querying the data
It looks that probably due to some updates (maybe doobie) event-log starts to fail when querying the data.
```
2020-08-13 15:40:31,690 ERROR application - Finding events to process failed
org.postgresql.util.PSQLException: Can't infer the SQL type to use for an instance of java.time.Instant. Use setObject() with an explicit Types value to specify the type to use.
```
This particular error happened on an environment created by a CI job but I believe it can be reproduced when deploying renku-graph `1.6.2` on any env.

### Reproduction steps

1. Deploy Renku with graph version 1.6.2 to a dev environment 
2.  in the renku project, go to charts/renku/requirements.yaml
3. In a project, add a dataset 

It would be great if we could make the test verifying the failing query (Finding events to process) to fail, too.

It could worth checking if maybe some other queries in the event-log and maybe projects-tokens service are not affected by the change.

Hint: these two could be relevant:
* https://github.com/tpolecat/doobie/issues/1095
* https://github.com/tpolecat/doobie/releases
Maybe we can simply extend the event-log's `TypesSerializers` with relevant traits so it becomes `trait TypesSerializers extends TimeMetaInstances with SqlMeta with MetaConstructors`?

travissarles:
In a previous PR, @joke1196  upgraded the Doobie library which uses JDBC under the hood to interact with Postgres. The upgrade changed the way in which java.time.Instant gets converted to a Postgres date type. We introduce new implicits to make the conversion work again. 

closes #354

### DIFF
--- a/event-log/src/main/scala/io/renku/eventlog/TypesSerializers.scala
+++ b/event-log/src/main/scala/io/renku/eventlog/TypesSerializers.scala
@@ -19,14 +19,15 @@
 package io.renku.eventlog
 
 import java.time.Instant
+
 import ch.datascience.graph.model.events.{BatchDate, CompoundEventId, EventBody, EventId}
 import ch.datascience.graph.model.projects
-import doobie.util.meta.{MetaConstructors, SqlMeta, TimeMetaInstances}
+import doobie.util.meta.{LegacyInstantMetaInstance, LegacyLocalDateMetaInstance}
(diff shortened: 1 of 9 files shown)


jachro (approved):
👍

PR #366, opened 2020-08-20
Fix 339 dataset keywords

joke1196:
Added proper index in Jena for schema:alternateName. Returning schema:keywords when querying dataset details

### DIFF
--- a/knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/BaseDetailsFinder.scala
+++ b/knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/BaseDetailsFinder.scala
@@ -21,11 +21,12 @@ package ch.datascience.knowledgegraph.datasets.rest
 import cats.MonadError
 import cats.effect.{ContextShift, IO, Timer}
 import ch.datascience.graph.config.RenkuBaseUrl
-import ch.datascience.graph.model.datasets.Identifier
+import ch.datascience.graph.model.datasets.{Identifier, Keyword}
 import ch.datascience.knowledgegraph.datasets.model.Dataset
 import ch.datascience.rdfstore._
 import eu.timepit.refined.auto._
 import io.chrisdavenport.log4cats.Logger
+import io.circe.Decoder.decodeList
 import io.circe.HCursor
 
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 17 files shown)


knowledge-graph/src/test/scala/ch/datascience/knowledgegraph/datasets/DatasetsGenerators.scala
```diff
@@ -41,10 +41,20 @@ object DatasetsGenerators {
       maybeUrl         <- Gen.option(datasetUrls)
       sameas           <- sameAs
       maybeDescription <- Gen.option(datasetDescriptions)
+      keywords         <- nonEmptyList(datasetKeywords)
```
jachro, line 44: I think keywords could potentially be empty so maybe generator should generate such cases, too.

knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/BaseDetailsFinder.scala
```diff
@@ -103,6 +104,59 @@ private class BaseDetailsFinder(
...
+    name = "ds by id - keyword details",
+    Set(
+      "PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>",
+      "PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>",
+      "PREFIX schema: <http://schema.org/>"
+    ),
+    s"""|SELECT DISTINCT ?datasetId ?keyword
+        |WHERE {
+        |  {
+        |    SELECT ?topmostSameAs
```
jachro, line 120: Why do we need this whole section with the topmostSameAs?

knowledge-graph/src/test/scala/ch/datascience/knowledgegraph/datasets/DatasetsGenerators.scala
joke1196: Yes good idea!

knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/BaseDetailsFinder.scala
joke1196: Oh yes for some reason I thought that this would gather all the keywords from all the versions of the dataset, but the identifier is enough to filter

jachro:
@joke1196 do you think that should go as another bugfix in the `1.6.x` branch? It's a bit difficult as it fixes the indexing issues and introduces a new feature. My impression is that the fixing nature of the change is more important here.

jachro (approved):
👍

jachro (approved):
👍

jachro (approved):
👍

PR #367, opened 2020-08-20
fix: so the publish job succeeds

jachro:
It looks that the publishing process cannot succeed on our CI. It fails with:
```
Error: no repository definition for https://charts.bitnami.com/bitnami. Please add them via 'helm repo add'
```
https://travis-ci.org/github/SwissDataScienceCenter/renku-graph/builds/719663096

This change tries to address that.

Reference: https://github.com/bitnami/charts


ableuler (approved):

jachro:
The fix got merged to #366. Closing.

PR #369, opened 2020-08-24
Db object metrics

### LINKED ISSUE #347: Metrics showing KG objects stats
As a graph-services owner, I'd like graph-services to send Prometheus metrics about the number of various types of objects in the Triples Store.

**Acceptance criteria:**
* decide on object types for the stats (`schema:Dataset`, `schema:Project`, `schema:ProcessRun` seem to be an obvious choice but probably these are not the only ones which are important);
* consider the performance of the query finding stats data so it does not affect other users;
* HINT: maybe `EventLogMetrics` could be used as an example.

joke1196:
closes #347

### DIFF
--- a/knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/Microservice.scala
+++ b/knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/Microservice.scala
@@ -83,13 +83,13 @@ object Microservice extends IOMicroservice {
 private class MicroserviceRunner(
     sentryInitializer:   SentryInitializer[IO],
     httpServer:          HttpServer[IO],
-    entitiesCountGauge:  KGMetrics[IO]
+    kgMetrics:           KGMetrics[IO]
 )(implicit contextShift: ContextShift[IO]) {
 
   def run(args: List[String]): IO[ExitCode] =
     for {
       _        <- sentryInitializer.run
-      _        <- entitiesCountGauge.run.start
+      _        <- kgMetrics.run.start
       exitCode <- httpServer.run
     } yield exitCode
 }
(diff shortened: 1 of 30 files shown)


knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/Microservice.scala
```diff
@@ -70,22 +73,23 @@ object Microservice extends IOMicroservice {
...
 
 }
 
 private class MicroserviceRunner(
-    sentryInitializer: SentryInitializer[IO],
-    httpServer:        HttpServer[IO]
-) {
+    sentryInitializer:   SentryInitializer[IO],
+    httpServer:          HttpServer[IO],
+    entitiesCountGauge:  KGMetrics[IO]
```
jachro, line 86: Maybe `entitiesCountGauge` could be named after the type?

knowledge-graph/src/test/scala/ch/datascience/knowledgegraph/metrics/MetricsGenerators.scala
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ch.datascience.knowledgegraph.metrics
+
+import ch.datascience.knowledgegraph.metrics.KGEntityType.{Dataset, ProcessRun, Project}
+import org.scalacheck.Gen
+
+object MetricsGenerators {
```
jachro, line 24: Maybe private?

knowledge-graph/src/test/scala/ch/datascience/knowledgegraph/metrics/MetricsGenerators.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  implicit val entitiesType: Gen[KGEntityType] = Gen.oneOf(
```
jachro, line 26: Maybe the `KGEntityType.all` can be used here so we don't miss any option?

jachro (approved):
👍

PR #424, opened 2020-09-28
fix: fix queries and tests for derivedFrom structure in KG

travissarles:
Signed-off-by: Travis Lee <travis.lee@sdsc.ethz.ch>


jachro (approved):
👍

jachro (approved):
👍

PR #446, opened 2020-10-08
fix: reduce number of threads required to track subscribers availability

### LINKED ISSUE #444: Event-log gets restarted
It looks there's some memory leakage which causes EL to restart.

Notice that the memory usage is gradually going up:
Screen Shot 2020-10-07 at 11 15 17

Hypothesis: Subscribers are getting freed up but there are too many of the objects which are freeing them. Each request has a thread. We fixed the thread pool. Cats IO creates fibers, which are a lightweight version of a thread. Looking at number of threads, all will always be used.

Instead we could just store the time when a subscriber should be freed up. Create a process which goes through subscribers to check for free ones.

travissarles:
closes #444

### DIFF
--- a/event-log/src/test/scala/io/renku/eventlog/subscriptions/SubscribersRegistrySpec.scala
+++ b/event-log/src/test/scala/io/renku/eventlog/subscriptions/SubscribersRegistrySpec.scala
@@ -24,9 +24,10 @@ import java.time.Instant
 import cats.effect.{ContextShift, IO, Timer}
 import cats.syntax.all._
 import ch.datascience.generators.Generators.Implicits.GenOps
-import ch.datascience.generators.Generators._
-import eu.timepit.refined.auto._
+import ch.datascience.generators.Generators.exceptions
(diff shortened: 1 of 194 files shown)


event-log/src/test/scala/io/renku/eventlog/subscriptions/SubscribersRegistrySpec.scala
```diff
@@ -0,0 +1,249 @@
...
+      subscribersRegistry.findAvailableSubscriber()          shouldBe Some(subscriberUrl)
+
+      val subscribersFound = (1 to 10).foldLeft(Set.empty[SubscriberUrl]) { (returnedSubscribers, _) =>
+        subscribersRegistry.findAvailableSubscriber().map(returnedSubscribers + _).getOrElse(returnedSubscribers)
+      }
+      subscribersFound shouldBe Set(subscriberUrl)
+    }
+  }
+
+  "start" should {
```
travissarles, line 195: Looks good

jachro (approved):
👍

PR #447, opened 2020-10-08
fix: escaping backslashes in user names

joke1196:


jachro:
O short one. Do you think it would make sense to run this kind of escaping mechanism on whole sparql queries, not the names only? I think here about some general rule similar to what we have in the `SparqlValueEncoder`. Maybe it should go there?

travissarles (dismissed):

travissarles (dismissed):

jachro (approved):
👍

PR #452, opened 2020-10-13
fix: classify GitLab 502 error & host resolution error as recoverable

travissarles:
Signed-off-by: Travis Lee <travis.lee@sdsc.ethz.ch>


jachro (approved):
👍

PR #460, opened 2020-10-15
update queries to exclude deleted datasets

joke1196:


knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/BaseDetailsFinder.scala
```diff
@@ -61,11 +61,23 @@ private class BaseDetailsFinder(
...
+        |               prov:atLocation ?location ;
         |               renku:topmostSameAs ?topmostSameAs;
         |               renku:topmostDerivedFrom/schema:identifier ?initialVersion.
+        |               
+        |    BIND(CONCAT(?location, "/metadata.yml") AS ?metaDataLocation) .
+        |    FILTER NOT EXISTS {
+        |      # Removing dataset that have an activity that invalidates them
+        |      ?deprecationEntity rdf:type <http://www.w3.org/ns/prov#Entity>;
+        |                         prov:atLocation ?metaDataLocation ;
+        |                         prov:wasInvalidatedBy ?invalidationActivity ;
```
travissarles, line 77: Do we use this invalidationActivity anywhere or is there an equivalent of _ so we can ignore it?

knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/DatasetsFinder.scala
```diff
@@ -118,15 +118,27 @@ private class IODatasetsFinder(
...
-        |          ?someId prov:wasDerivedFrom/schema:url ?dsId.
-        |          ?someId schema:isPartOf ?projectId.
+        |            # Removing dataset that have an activity that invalidates them
+        |            ?deprecationEntity rdf:type <http://www.w3.org/ns/prov#Entity>;
+        |                               prov:atLocation ?metaDataLocation ;
+        |                               prov:wasInvalidatedBy ?invalidationActivity ;
+        |                               schema:isPartOf ?projectId .
+        |        }
+        |        
+        |        FILTER NOT EXISTS {
```
travissarles, line 139: Did you determine that these two filter clauses are executed together or sequentially? Does it matter?

travissarles (commented):
Do we need a case for when the original datasset was deleted by one project a forked project is still using it?

knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/BaseDetailsFinder.scala
joke1196: Yes this means that I want the prov:wasInvalidatedBy to be pointing to something

knowledge-graph/src/main/scala/ch/datascience/knowledgegraph/datasets/rest/DatasetsFinder.scala
joke1196: It could matter for performance, they should be executed sequentially. This is some kind of an OR statement so the order does not really matter. We want to filter out the dataset with an invalidation activity or the dataset that are derivedFrom.

travissarles (approved):

PR #461, opened 2020-10-16
Memory leak on event-log

jachro:
There's a tiny problem which causes huge memory leak on event-log. The SubscribersRegistry creates a new thread each time it checks if there are subscribers in the busy pool.

### DIFF
--- a/event-log/src/main/scala/io/renku/eventlog/subscriptions/SubscribersRegistry.scala
+++ b/event-log/src/main/scala/io/renku/eventlog/subscriptions/SubscribersRegistry.scala
@@ -102,14 +102,12 @@ private class SubscribersRegistry(
 
   def subscriberCount(): Int = busyPool.size() + availablePool.size()
 
-  private def start(): IO[Fiber[IO, Unit]] = {
-    for {
-      _                        <- timer sleep checkupInterval
-      subscribersDueForCheckup <- findSubscribersDueForCheckup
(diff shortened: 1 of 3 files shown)


event-log/src/main/scala/io/renku/eventlog/subscriptions/SubscribersRegistry.scala
```diff
@@ -102,14 +102,12 @@ private class SubscribersRegistry(
...
-      subscribersDueForCheckup <- findSubscribersDueForCheckup
-      _                        <- bringToAvailable(subscribersDueForCheckup)
-      _                        <- start()
-    } yield ()
-  }.start
+  private def busySubscriberCheckup(): IO[Unit] = for {
+    _                        <- timer sleep checkupInterval
+    subscribersDueForCheckup <- findSubscribersDueForCheckup
+    _                        <- bringToAvailable(subscribersDueForCheckup)
+    _                        <- busySubscriberCheckup()
```
travissarles, line 109: What happens if this method gets called from different threads? Also, could we start this method name with a verb? checkOnBusySubscribers or something.

travissarles (approved):

event-log/src/main/scala/io/renku/eventlog/subscriptions/SubscribersRegistry.scala
jachro: As it's a private method I think we've got control when it's called.

PR #488, opened 2020-11-03
fix: dirty repo on checkout when using renku migrate

joke1196:
## The issue
Checking out a specific commit makes the repo dirty with files having changed. This will happen on repos tracking files with `git lfs`
Git lfs figures out, with the help of the rules in the `.gitattributes` file, which files should be tracked. 
If, in this particular commit, some files should be in git lfs but are not; they will be **automatically** added. This means that a change occurred on this files and these changes will be tracked by `git`, so the repo now contains changes even though we just checked out a commit.

## The solution

For our use case the fix is to remove the .gitattributes file and checkout out the current commit:
```
git rm .gitattributes
git checkout .
```


triples-generator/src/main/scala/ch/datascience/triplesgenerator/eventprocessing/triplesgeneration/renkulog/Commands.scala
```diff
@@ -96,6 +103,11 @@ private object Commands {
         %%("git", "checkout", commitId.value)(repositoryDirectory)
       }.map(_ => ())
 
+    def checkoutCurrent(repositoryDirectory: Path): IO[Unit] =
+      IO {
+        %%("git", "checkout", ".")(repositoryDirectory)
+      }.map(_ => ())
```
jachro, line 109: Could you maybe replace all `map(_ => ())` with `void`. I guess just in this file :)

triples-generator/src/main/scala/ch/datascience/triplesgenerator/eventprocessing/triplesgeneration/renkulog/RenkuLogTriplesGenerator.scala
```diff
@@ -82,6 +86,18 @@ private[eventprocessing] class RenkuLogTriplesGenerator private[renkulog] (
       _             <- (git.checkout(commitEvent.commitId, repoDirectory)).toRight
     } yield ()
 
+  private def cleanUpRepository(repoDirectory: Path) = {
+    val gitAttributeFilePath: Path = repoDirectory / gitAttributeFileName
+    file.exists(gitAttributeFilePath).flatMap {
+      case false => IO.unit
+      case true =>
+        for {
```
jachro, line 94: I guess it would be nice if `repoDirectory` is implicit by I think it's not a simple change...

jachro (dismissed):
👍

rokroskar:
Can you describe the solution? this happens sometimes in interactive sessions as well so maybe it would be good to have a general fix. To make sure we're on the same page, do you have the description of the problem somewhere?

jachro (dismissed):
👍

jachro (approved):
👍

PR #494, opened 2020-11-04
232 flatten json ld

### LINKED ISSUE #232: JsonLD to allow flattening
As a Renku dev, I'd like the JsonLD lib to allow flattening of the nested JsonLD structure.

Acceptance criteria:
* in case of non-nested JsonLD object the flatten operation should return the same object;
* in case of nested JsonLD object the flatten operation should return a JsonLD array of all the objects found in the structure; the nested objects should be replaced with their ids;
* in case of an array of nested JsonLD objects, the operation mentioned in the points above, should be applied to all the objects in the array.

Some work for that has been done and pushed to the `232-json-flatten` branch.

### LINKED ISSUE #333: Relax constraints on JsonLDEntities to handle List instead of NonEmptyList
Acceptance criteria:
 
 * the serialization of JsonLDEntities without a list of properties should be successful and return an object with the `@id`

travissarles:
closes #232 
closes #333


json-ld/src/test/scala/io/renku/jsonld/JsonLDSpec.scala
```diff
@@ -455,6 +460,246 @@ class JsonLDSpec extends AnyWordSpec with ScalaCheckPropertyChecks with should.M
     }
   }
 
+  "flatten" should {
```
joke1196, line 463: I think we can get rid of these tests as flatten was moved to the Flattener class

json-ld/src/main/scala/io/renku/jsonld/JsonLD.scala
```diff
@@ -139,14 +194,67 @@ object JsonLD {
...
-    override lazy val entityTypes: Option[EntityTypes] = None
+    override lazy val toJson:      Json                   = Json.obj("@id" -> id.asJson)
+    override lazy val entityId:    Option[EntityId]       = None
+    override lazy val entityTypes: Option[EntityTypes]    = None
+    override lazy val asArray:     Option[Vector[JsonLD]] = Some(Vector(this))
+  }
+
+  final case class MalformedJsonLD(message: String) extends RuntimeException(message)
+
+  private def deNestProperties(entity:                  JsonLDEntity,
```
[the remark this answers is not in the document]
travissarles: nope!

jachro:
I've just realised that we forgot about one thing :) To make use of the `flatten` method. I believe we'd need to go through the code and look for places where we put together some JsonLD objects and then send them to Jena in the tests. An obvious place is `ch.datascience.rdfstore.InMemoryRdfStore#loadToStore` but I imagine there could be few more.

json-ld/src/main/scala/io/renku/jsonld/Flattener.scala
```diff
@@ -0,0 +1,112 @@
...
+ */
+
+package io.renku.jsonld
+
+import cats.syntax.all._
+import io.renku.jsonld.JsonLD.{JsonLDArray, JsonLDEntity, JsonLDEntityId, MalformedJsonLD}
+
+import scala.annotation.tailrec
+
+trait JsonLDArrayFlattener extends Flattener {
```
jachro, line 26: Would that maybe work if that trait is named `JsonLDArrayFlatten`. I think it'd kind of match better the functionality of the trait that it introduces the `flatten` method.

json-ld/src/main/scala/io/renku/jsonld/Flattener.scala
```diff
@@ -0,0 +1,112 @@
...
+                                jsons    <- deNest(List(jsonLDEntity), List.empty[JsonLDEntity])
+                                accRight <- acc
+                              } yield accRight ++ jsons
+                            case (acc, other) => acc.map(other +: _)
+                          }
+      flattenedArray <- checkForUniqueIds(flattenedJsons.distinct)
+    } yield flattenedArray
+}
+
+trait JsonLDEntityFlattener extends Flattener {
```
jachro, line 43: Similarly to the comment above about the trait name.

json-ld/src/main/scala/io/renku/jsonld/Flattener.scala
```diff
@@ -0,0 +1,112 @@
...
+  self: JsonLDEntity =>
+
+  lazy val flatten: Either[MalformedJsonLD, JsonLD] = deNest(List(this), Nil).flatMap(checkForUniqueIds)
+}
+
+private[jsonld] trait Flattener {
```
jachro, line 49: What would you think if we create a new 'flatten' package for these three traits? Then, effectively we could split this file into three (or four) files for each trait.

I think there could be some advantages of that:
* smaller compilation units;
* the `deNest` method could be `protected[flatten]` scope;
* the `checkForUniqueIds ` method could be `protected[flatten]` scope;

By the way, maybe the `checkForUniqueIds` could go into a separate trait like `Validation` in the `flatten` package?

travissarles (commented):
lgtm!

graph-commons/src/test/scala/ch/datascience/rdfstore/entities/InvalidationEntity.scala
```diff
@@ -0,0 +1,67 @@
...
+import ch.datascience.graph.model.events.CommitId
+import ch.datascience.rdfstore.FusekiBaseUrl
+import io.renku.jsonld.syntax._
+import io.renku.jsonld.{EntityId, JsonLDEncoder}
+
+final case class InvalidationEntity(
+    override val commitId:                  CommitId,
+    entityIdToInvalidate:                   String,
+    override val project:                   Project,
+    override val maybeInvalidationActivity: Option[Activity]
```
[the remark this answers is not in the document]
joke1196: We have to fit to the type on the Entity. So the InvalidationEntity is not a proper type, it's just an entity, so we cannot change the type of the properties on Entity. In the end InvalidationEntity is just to make it easier for us.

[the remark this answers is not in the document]
joke1196: Oh got what you were saying I can default it to some sure

joke1196 (approved):

PR #508, opened 2020-11-12
tg and el stop working at some circumstances

### LINKED ISSUE #507: Event-log dispatch and triples-generator subscribe stop working on failure
It looks that event-log's dispatch process does not resume after postgres outage.
Similarly, triples-generator subscription mechanism stop re-subscribing after event-log being not available longer than the number of retries TG performs.

A log statement after which EL's dispatch stopped pulling new events:
```
ERROR application - Finding events to dispatch failed
java.sql.SQLTransientConnectionException: HikariPool-1 - Connection is not available, request timed out after 30002ms.```

### PR DESCRIPTION
**Pull request by jachro**
closes #507

### DIFF
--- a/event-log/src/main/scala/io/renku/eventlog/subscriptions/EventsDispatcher.scala
+++ b/event-log/src/main/scala/io/renku/eventlog/subscriptions/EventsDispatcher.scala
@@ -61,7 +61,7 @@ class EventsDispatcher(
   private def popEvent: IO[(CompoundEventId, EventBody)] =
     eventsFinder
       .popEvent()
-      .recoverWith(loggingError(retry = eventsFinder.popEvent()))
+      .recoverWith(loggingErrorAndRetry(retry = eventsFinder.popEvent))
       .flatMap {
         case None            => (timer sleep noEventSleep) flatMap (_ => popEvent)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: event-log/src/main/scala/io/renku/eventlog/subscriptions/EventsDispatcher.scala [hunk 1]
```diff
@@ -109,12 +109,13 @@ class EventsDispatcher(
...
-      _      <- logger.error(exception)("Finding events to dispatch failed")
-      _      <- timer sleep onErrorSleep
-      result <- retry
-    } yield result
+  private def loggingErrorAndRetry[O](retry: () => IO[O]): PartialFunction[Throwable, IO[O]] = {
+    case NonFatal(exception) =>
+      for {
+        _      <- logger.error(exception)("Finding events to dispatch failed")
+        _      <- timer sleep onErrorSleep
+        result <- retry() recoverWith loggingErrorAndRetry(retry)
```
joke1196, line 117: So is it ok if we retry indefinitely?

jachro: I believe so as for instance postgres could be down for some time but eventually should come back

joke1196 (approved):

PR #509, opened 2020-11-13
Graph services to load client certificates if configured

### LINKED ISSUE #432: graph services to work with CA-signed certificates
As a Renku maintainer, I'd like graph services to work with CA-signed certificates so they are trusted on https requests.

**Acceptance criteria:**
- [ ] investigate how graph services can follow the solution introduced for Python services where:
  * ~a CA-signed certificate is put into a config map and mounted to the service as a volume;~
  * ~an environment variable pointing to the cert is defined so it might be used by the service;~
  * the certificate is put into a secret and available for the services as an environment variable;
    - [ ] find a name for the secret (e.g. `CA-signed-certificate-secret`) and agree on it with the YAT team;
- [ ] find out what's the format of the certificate (the YAT team tries with `.pem` files - more on that [here](https://blog.codecentric.de/en/2013/01/how-to-use-self-signed-pem-client-certificates-in-java/));
- [ ] determine best method for adding cert to java trust store:

1. *Programmatic (Scala)*:  all microservices use the run() method so we could do add the logic in there. (see Sentry Initialisation process));
  * for programmatic [this](https://stackoverflow.com/questions/21775408/programmatically-add-a-self-signed-certificate-to-your-keystore-truststore) could be helpful;
2. *Bash script in config map*
  * for bash script/config map [this](https://stackoverflow.com/questions/11617210/how-to-properly-import-a-selfsigned-certificate-into-java-keystore-that-is-avail) might be useful; (there's a `lucene-config.yaml` in `renku-graph` and a `configmap.yaml` in the `renku` project as examples)
- [ ] find out if adding the cert to the trust store is enough for http4s clients to successfully use it;
- [ ] find out a way of how the set-up might be tested;

More info, also about testing, can be found [here](https://connect2id.com/products/nimbus-oauth-openid-connect-sdk/examples/utils/custom-trust-store).

openssl req -x509 -new -nodes -key myCA.key -sha256 -days 1825 -out myCA.pem

## Questions
- Which services need to use this cert? The ones which contact GitLab?

jachro:
closes #432

### DIFF
--- a/acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/db/EventLog.scala
+++ b/acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/db/EventLog.scala
@@ -21,6 +21,7 @@ package ch.datascience.graph.acceptancetests.db
 import cats.data.NonEmptyList
 import cats.effect.{ContextShift, IO}
 import ch.datascience.db.{DBConfigProvider, DbTransactor}
+import ch.datascience.graph.acceptancetests.tooling.TestLogger
 import ch.datascience.graph.model.events.{CommitId, EventId}
 import ch.datascience.graph.model.projects.Id
 import com.dimafeng.testcontainers.{Container, JdbcDatabaseContainer, PostgreSQLContainer}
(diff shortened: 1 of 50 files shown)


travissarles (approved):

travissarles (dismissed):

pameladelgado (approved):
Thank you @jachro very much for this! :raised_hands: 
The certificate-related changes look good to me. In my opinion it makes sense to add the certificate to the global section like you did, so that later other components can support CA certificates properly too.
This change is also harmless to other existing deployment configurations so we can safely merge this.

PR #512, opened 2020-11-19
feat: allowing re-association of token to public projects

### LINKED ISSUE #506: Webhook's POST /projects/:id/webhooks/validation to refresh tokens for public projects
It looks that if a project is public and user's token gets invalidated, the `POST /projects/:id/webhooks/validation` does not refresh user's token stored in the tokens-repository with the one coming with the request to the validation endpoint, even if it's valid. This mechanism seems to be in place for non-public projects.

joke1196:
closes #506


jachro (approved):
Cool! Let's merge it!

PR #518, opened 2020-11-20
chore: added cpu limits and requests to TGs values.yaml

### LINKED ISSUE #516: Define default CPU requests/limits
It would be nice to define a default CPU limit (and potentially a request) to match the defined memory limit for the pods.

joke1196:
closes #516

### DIFF
--- a/helm-chart/renku-graph/values.yaml
+++ b/helm-chart/renku-graph/values.yaml
@@ -33,6 +33,9 @@ triplesGenerator:
   resources:
     limits:
       memory: 5000Mi
+      cpu: "2"
+    requests:
+      cpu: "1"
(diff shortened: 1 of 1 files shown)


jachro:
Actually, after the discussion with Andreas and Pamela I'm wondering if we still need to set the limits/requests. The reason I'm not sure is that we need to tweak the values anyway on per deployment settings and the defaults are effectively what we get without setting them explicitly.
Any thoughts?

joke1196:
Yes sure we can drop it here the most important one was the production. We could leave it as is for DEV

jachro:
As it looks the CPU settings need to be set on per deployment basis, we don't have to default them on the graph side.

PR #534, opened 2020-11-26
533 subscription category registry

### LINKED ISSUE #425: Event-log to allow subscription to different statuses
As a renku-graph maintainer, I'd like event-log's subscription mechanism to allow subscribing for different event statuses.

See statuses in the [README](https://github.com/SwissDataScienceCenter/renku-graph/blob/master/event-log/README.md#post-subscriptions)

### Current design
The [EventFetcher](https://github.com/SwissDataScienceCenter/renku-graph/blob/master/event-log/src/main/scala/io/renku/eventlog/subscriptions/EventFetcher.scala#L85) in event-log looks for New or Recoverable events. We should parameterize this. 

### Possible implementation 
We need to create multiple implementations of `EventsDispatcher`, `SubscribersRegistry`, and `Subscriptions` to handle multiple statuses (e.g. `GeneratedTriples`) or perhaps other properties besides statuses. We could have a `Governor` to control multiple instances of these classes. In the future there will be a new property on the event called `lastChecked`. `EventFetcher` could be a trait.

**Acceptance criteria:**
- [ ] modify the behaviour of the `POST /subscriptions` endpoint so subscriptions to statuses specified in the payload are used and not internally fixed to `NEW` and `RECOVERABLE_FAILURE`;
  - [ ] introduce a mechanism managing subscription for various statuses which would be:
    - [ ] instantiating a new Event Dispatcher with Subscriptions when there's no dispatcher for statuses subscription is requested through the endpoint;
    - [ ] adding subscriber to Subscriptions for statuses Event Dispatcher already exists;
    - [ ] making sure Event Dispatchers for each group of statuses run concurrently and independently;
    - [ ] making sure Event Dispatchers for each group have their own instance of the `Subscription` class;
    - [ ] making sure Event Dispatcher for each group updates polls for events of relevant statuses before it dequeues them;
    - [ ] making sure Event Dispatcher for each group updates the event status of the dequeued event;

**Steps:**
* renaming of the current classes / pulling fetcher to a separate package;
* create an EventDeserializer;
* create the SubscriptionType class;
* create the SubscriptionsTypesRegistry class;
events-dispatching-rework

travissarles:
closes #425


event-log/src/main/scala/io/renku/eventlog/subscriptions/SubscriptionCategoryRegistry.scala
```diff
@@ -0,0 +1,83 @@
...
+) extends SubscriptionCategoryRegistry[Interpretation] {
+  override def run(): Interpretation[Unit] = categories.toList.traverse_(_.run())
+
+  override def register(subscriptionRequest: Json): Interpretation[Either[RequestError, Unit]] =
+    if (categories.isEmpty) {
+      Either.left[RequestError, Unit](NoCategoriesAvailable).pure[Interpretation]
+    } else {
+      categories.toList
+        .map(_.register(subscriptionRequest).map {
+          case Some(_) => Some(())
```
joke1196, line 51: Not totally convinced about this part there may be a better way of doing this. Not sure why but sequence does not understand the abstract type SubscriptionCategory#PayloadType. There could be an implicit missing

event-log/src/main/scala/io/renku/eventlog/subscriptions/EventsDistributor.scala
```diff
@@ -35,11 +35,11 @@ import scala.concurrent.duration._
 import scala.language.postfixOps
 import scala.util.control.NonFatal
 
-trait EventsDistributor[Interpretation[_]] {
+private[subscriptions] trait EventsDistributor[Interpretation[_]] {
```
jachro, line 38: Oh, shouldn't that be simply `private`?

event-log/src/main/scala/io/renku/eventlog/subscriptions/SubscriptionCategory.scala
```diff
@@ -19,41 +19,27 @@
...
+private trait SubscriptionCategory[Interpretation[_]] {
   def run(): Interpretation[Unit]
 
-  def register(payload: Json): Interpretation[Option[T]]
+  def register(payload: Json): Interpretation[RegistrationResult]
+
 }
 
-class SubscriptionCategoryImpl[Interpretation[_]: Effect, T <: SubscriptionCategoryPayload](
+private[subscriptions] class SubscriptionCategoryImpl[Interpretation[_]: Effect, T <: SubscriptionCategoryPayload](
```
jachro, line 32: Maybe just `private` here, too?

event-log/src/main/scala/io/renku/eventlog/subscriptions/SubscriptionCategoryRegistry.scala
```diff
@@ -0,0 +1,75 @@
...
+import scala.concurrent.ExecutionContext
+
+trait SubscriptionCategoryRegistry[Interpretation[_]] {
+
+  def run(): Interpretation[Unit]
+
+  def register(subscriptionRequest: Json): Interpretation[SubscriptionResult]
+}
+
+private[subscriptions] class SubscriptionCategoryRegistryImpl[Interpretation[_]: Effect: Applicative](
```
jachro, line 40: Another `private` potentially?

event-log/src/main/scala/io/renku/eventlog/subscriptions/package.scala
```diff
@@ -35,4 +36,19 @@ package object subscriptions {
   trait SubscriptionCategoryPayload {
     def subscriberUrl: SubscriberUrl
   }
+
+  sealed trait SubscriptionResult
```
jachro, line 40: I guess as a good practice the subtypes of the `SubscriptionResult` should be nested in the `SubscriptionResult` companion. And, what do you think if the type goes to companion of the `SubscriptionCategory`?

event-log/src/main/scala/io/renku/eventlog/subscriptions/package.scala
```diff
@@ -35,4 +36,19 @@ package object subscriptions {
...
+  final case object SuccesfulSubscription extends SubscriptionResult
+  final case class UnsupportedPayload(message: String) extends SubscriptionResult
+  final case object NoCategoriesAvailable extends SubscriptionResult
+
+  sealed trait RegistrationResult
```
jachro, line 45: Probably we could do a similar thing as for the `SubscriptionResult`?

travissarles:
lgtm

event-log/src/main/scala/io/renku/eventlog/subscriptions/package.scala
joke1196: So SubscriptionResult I will put in SubscriptionCategoryRegistry as this is the type returned to the endpoint. And I'll put RegistrationResult in the SubscriptionCategory

jachro (approved):

PR #535, opened 2020-11-27
Replace view with project table on the event_log DB

### LINKED ISSUE #526: Project table on the event-log DB
As a renku-graph maintainer, I'd like the `project_latest_event_date` is replaced with a `projects` table.

**The problem**
It looks that the `project_latest_event_date` materialized view does the job in terms of logic but not in terms of performance. The reason for that is that it is refreshed every single time an even is popped from the `event_log` table. It looks more reasonable to replace the view with a regular table which would get eventually update when needed.

**Acceptance criteria**
- [x] create a new DB migration process which would create a new `project` table with `project_id`, `project_path` and `latest_event_date` columns;
  - [x] `project_id` to be primary key;
  - [x] indices to be created for `project_path` and `latest_event_date`;
- [x] create a new DB migration process which would fill in the table using the same query as in the `project_latest_event_date` view;
  - [x] the migration to be run only if the `project` table is empty to prevent from running it on every event-log service start-up;
- [x] make the flow which adds new events to update the new `project`'s `latest_event_date` if a new event's `event_date` is later than the date currently stored for that event project;
- [x] update all the queries using the `project_latest_event_date` view so the new `project` table is used instead;
- [x] create a new DB migration process removing the `project_path` from the `event_log` table;
- [x] create a new DB migration process removing the `project_latest_event_date` materialized view.

jachro:
closes #526


travissarles (approved):
awesome!

PR #538, opened 2020-12-01
migration of event status from processing to generating_triples

joke1196:


event-log/src/test/scala/io/renku/eventlog/init/EventStatusRenamerImplSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  private trait TestCase {
+    val logger             = TestLogger[IO]()
+    val eventStatusRenamer = new EventStatusRenamerImpl[IO](transactor, logger)
+  }
+
+  private def store(event: Event, withStatus: String): Unit = {
+    upsertProject(event.compoundEventId, event.project.path, event.date)
+    execute {
```
joke1196, line 90: Sadly I could not use the method in the EventLogDataProvisioning trait as I removed the Processing EventStatus, I needed to pass in a String

jachro (approved):
👍

PR #543, opened 2020-12-03
422 new event statuses

joke1196:
This issue is part of issue 422:
The goal is to add new event statuses that will be used later on by the TG and EL, to have a better understanding of what is currently happening to the event and also to facilitate the transition to renku core when generating triples.
Theses new statuses also have to be reflected in the metrics.


jachro (commented):
That looks really good. I really like the names of the gauges.
I'm just wondering if we'd maybe like to keep JSONs for the dashboards somewhere in the renku-graph project. Maybe in some `dashboards` folder? What do you think?

jachro (approved):

PR #544, opened 2020-12-03
refactor: single CI build on PR merge

### LINKED ISSUE #500: Simplify KG release process
As a renku-graph maintainer, I'd like the KG release process is a bit simpler so it doesn't produce unnecessary commits (e.g. for snapshots).

**The problem**
Now, once a PR gets merged to master the following steps happens:
* CI build from the merge commit which produces two commits: one for the new release and one for the new snapshot;
* CI build for the new release (it creates new images and a helm chart);
* CI build for the new snapshot.

**Eventual solution**
It looks the process could be streamlined a lot. What if we put all the steps into a single job executed during merge to master? The flow would change and should look more or less like follows:
* when a PR gets merged to master, the only CI build would:
  * test;
  * tag (take values from the `version.sbt` as is;
  * push the tag;
  * run `chartpress`;
* the build happening for the commit where setting the next version or tag would go;
As a result, the commit history in master would be very slim and would contain just the merges to master.

**Drawbacks** 
* the release plugin would need to be customised even more than it's now;
* in the process of development, when a new issue is picked-up for work, probably the first operation for a dev would be to decide about the release version and change both the `version.sbt` and `Chart.yaml` files.

jachro:
closes #500


.github/workflows/test-release-publish.yml
```diff
@@ -1,8 +1,14 @@
-name: Test and publish
+name: Test and release
 on:
   push:
-    tags:
-      - '*.*.*'
+    branches:
+      - master
+      - 'hotfix-[0-9]+.[0-9]+**'
```
travissarles, line 6: do we use this hotfix tag?

travissarles: or should we :D?

webhook-service/src/test/scala/ch/datascience/webhookservice/eventprocessing/IOProcessingStatusFetcherSpec.scala
```diff
@@ -160,7 +160,7 @@ class IOProcessingStatusFetcherSpec
     }
 
     "fail if total < done" in {
-      val done  = nonNegativeInts().generateOne.value + 1
```
travissarles, line 163: is this related?

.github/workflows/test-release-publish.yml
joke1196: I think this goes away as now we will tag the release and push the image to the registry. We will remove this second part where we create another commit for the tag.

joke1196 (approved):

PR #548, opened 2020-12-04
fix: single CI job to publish tag

jachro:


joke1196:
Looks good. I was wondering if we needed on line 90 of the publishing the:
- uses: actions/checkout@v2
        with:
          fetch-depth: 0

jachro:
Yes, not so sure

joke1196 (approved):

PR #551, opened 2020-12-08
422 event payload table

joke1196:
This PR creates and confirgure a new `event_payload` table in the `event_log`  DB


travissarles (approved):

event-log/src/main/scala/io/renku/eventlog/init/EventPayloadTableCreator.scala
```diff
@@ -0,0 +1,83 @@
...
+    extends EventPayloadTableCreator[Interpretation]
+    with EventTableCheck[Interpretation] {
+
+  import cats.syntax.all._
+  import doobie.implicits._
+
+  private implicit val transact: DbTransactor[Interpretation, EventLogDB] = transactor
+
+  override def run(): Interpretation[Unit] =
+    whenEventTableExists(
```
jachro, line 52: Shouldn't the new table actually be created only if the `event` table exists?

event-log/src/main/scala/io/renku/eventlog/init/EventPayloadTableCreator.scala
```diff
@@ -0,0 +1,83 @@
...
+      .recover { case _ => false }
+
+  private def createTable = for {
+    _ <- createTableSql.update.run transact transactor.get
+    _ <- execute(sql"CREATE INDEX IF NOT EXISTS idx_event_id ON event_payload(event_id)")
+    _ <- execute(sql"CREATE INDEX IF NOT EXISTS idx_project_id ON event_payload(project_id)")
+    _ <- logger info "'event_payload' table created"
+  } yield ()
+
+  private lazy val createTableSql = sql"""
```
jachro, line 75: I guess it would be nice if we have a foreign key on the table added. You can maybe take a look here:
https://github.com/SwissDataScienceCenter/renku-graph/blob/2a680e9db8859152962f0ed71a4992377e6acb60/event-log/src/main/scala/io/renku/eventlog/init/ProjectTableCreator.scala#L102
However, the foreign key will be probably also compound.

event-log/src/main/scala/io/renku/eventlog/init/DbInitializer.scala
```diff
@@ -46,6 +47,7 @@ class DbInitializerImpl[Interpretation[_]](
   override def run(): Interpretation[Unit] = {
     for {
       _ <- eventLogTableCreator.run()
+      _ <- eventPayloadTableCreator.run()
```
jachro, line 50: Wouldn't it be better to add as the very last step. That would reflect the history of event-log DB evolution and I think should make things simpler

event-log/src/main/scala/io/renku/eventlog/init/EventPayloadTableCreator.scala
joke1196: Yes good catch!

event-log/src/test/scala/io/renku/eventlog/init/EventPayloadTableCreatorSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+import cats.effect.IO
+import ch.datascience.interpreters.TestLogger
+import ch.datascience.interpreters.TestLogger.Level.Info
+import doobie.implicits._
+import org.scalatest.matchers.should
+import org.scalatest.wordspec.AnyWordSpec
+
+class EventPayloadTableCreatorSpec extends AnyWordSpec with DbInitSpec with should.Matchers {
+
+  protected override lazy val migrationsToRun: List[Migration] = List(eventLogTableCreator)
```
jachro, line 30: Wouldn't it better (from the migration specs consistency point of view) that we run all the migrations here (except the one we test here)?

event-log/src/test/scala/io/renku/eventlog/init/EventPayloadTableCreatorSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  "run" should {
+    "fail if the 'event' table does not exist" in new TestCase {
+
+      tableExists("event")         shouldBe false
```
jachro, line 35: If we'd run all the previous migrations, we'd probably need to use `dropTable('event')` here.

event-log/src/test/scala/io/renku/eventlog/init/EventPayloadTableCreatorSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+
+      intercept[Exception] {
+
+        tableCreator.run().unsafeRunSync()
+      }.getMessage shouldBe "Event table missing; creation of event_payload is not possible"
+    }
+
+    "create the 'event_payload' table if it doesn't exist" in new TestCase {
+
+      createEventTable()
```
jachro, line 46: That could be probably skipped (if the above is done)

event-log/src/test/scala/io/renku/eventlog/init/EventPayloadTableCreatorSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+      tableCreator.run().unsafeRunSync() shouldBe ((): Unit)
+
+      tableExists("event_payload") shouldBe true
+
+      logger.loggedOnly(Info("'event_payload' table created"))
+    }
+
+    "do nothing if the 'event_payload' table already exists" in new TestCase {
+
+      createEventTable()
```
jachro, line 58: The same as above

event-log/src/test/scala/io/renku/eventlog/init/EventPayloadTableCreatorSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+      logger.reset()
+
+      tableCreator.run().unsafeRunSync() shouldBe ((): Unit)
+
+      logger.loggedOnly(Info("'event_payload' table exists"))
+    }
+
+    "create indices for certain columns" in new TestCase {
+
+      createEventTable()
```
jachro, line 76: As above

jachro (approved):

PR #564, opened 2020-12-15
540 tg reprov decision to incorporate schema version

### LINKED ISSUE #540: TG to read info about schema version -> CLI version from the config
As a renku-graph maintainer, I'd like Renku CLI compatibility matrix to be introduced to the TG.

**The problem**
Currently, TG supports a single version of Renku CLI and a moreover, takes the decision about starting the re-provisioning process using the CLI version rather than schema version. That causes 
* the very costly re-provisioning process to be kicked-off on every CLI release (which in most cases does not change anything from the schema perspective);
* prevents us from bumping CLI version on TG as any new release arrives;

**The solution**
As we're moving towards the _No downtime re-provisioning_, we'd like to switch TG to use schema version instead of the CLI version. However, it seems reasonable that a compatibility matrix to replace a single CLI version support is introduced at that stage so we could both work with different CLI version producing the same schemas and to work with different schemas at the same time (in the future). TG can use also this info and when deciding about the re-provisioning process.

As part of the change the following compatibility matrix to be introduced in the TG's config:
* case when each item introduces schema version change:
  ```
  compatibility-matrix: [
    '0.13.1' -> 14,
    '0.12.7' -> 13
  ]
  ```
* case when there are CLI versions producing the same schemas but requiring reprovisioning anyway (e.g. when they introduce fixes to the JSON-LD payload generation):
   ```
   compatibility-matrix: [
     '0.13.2' -> 13,
     '0.13.1' -> 13
  ]
  ```

**Acceptance criteria:**
- [x] introduce the `compatibility-matrix` config property remembering that the order of the items matters;
- [x] introduce HELM chart and values changes providing value for the new config property;
- [x] change the current re-provisioning mechanism so the new `renku:SchemaVersion`object is added on top of the `renku:CliVersion` (potentially a new object type is created containing both the versions);
- [x] TG to treat the first item in the matrix as the most recent version of the schema;
- [x] TG to check on start-up if the installed version of CLI matches the most recent version specified in the matrix;
- [x] TG to read the most recent version of the schema;
- [x] TG to kick-off re-provisioning if:
  * the most recent schema version does not match the schema version in the TS (lack of schema version in the TS mean no match);
  * the most recent schema version matches the schema version in the TS and
    * there are two items if the matrix with the same schema version;
    * CLI version in the TS does not match the version of the CLI installed on the TG;
- [x] TG to update both the `schemaVersion` and the `CLIVersion` in the TS;
- [x] TG to check if `RENKU_PYTHON_DEV_VERSION` env variable is set (or empty or containing whitespaces) and never kick-off re-provisioning if its value is non-blank;
  - [x] TG to log a warning if running on the CLI DEV version.

travissarles:
closes #540


triples-generator/src/main/scala/ch/datascience/triplesgenerator/models/package.scala
```diff
@@ -0,0 +1,43 @@
...
+package ch.datascience.triplesgenerator
+
+import ch.datascience.graph.model.CliVersion
+import ch.datascience.graph.model.projects.SchemaVersion
+import io.circe.Decoder
+import io.circe.Decoder.decodeList
+
+package object models {
+  final case class RenkuVersionPair(cliVersion: CliVersion, schemaVersion: SchemaVersion)
+      extends Product
```
joke1196, line 28: Already defined by case class

triples-generator/src/main/scala/ch/datascience/triplesgenerator/models/package.scala
```diff
@@ -0,0 +1,43 @@
...
+
+package ch.datascience.triplesgenerator
+
+import ch.datascience.graph.model.CliVersion
+import ch.datascience.graph.model.projects.SchemaVersion
+import io.circe.Decoder
+import io.circe.Decoder.decodeList
+
+package object models {
+  final case class RenkuVersionPair(cliVersion: CliVersion, schemaVersion: SchemaVersion)
```
joke1196, line 27: Move to model package (same as CliVersion)

triples-generator/src/main/scala/ch/datascience/triplesgenerator/Microservice.scala
```diff
@@ -54,20 +57,34 @@ object Microservice extends IOMicroservice {
...
-      sparqlTimeRecorder       <- SparqlQueryTimeRecorder(metricsRegistry)
-      reProvisioningStatus     <- ReProvisioningStatus(subscriber, ApplicationLogger, sparqlTimeRecorder)
-      reProvisioning           <- IOReProvisioning(triplesGeneration, reProvisioningStatus, sparqlTimeRecorder, ApplicationLogger)
+      certificateLoader          <- CertificateLoader[IO](ApplicationLogger)
+      gitCertificateInstaller    <- GitCertificateInstaller[IO](ApplicationLogger)
+      fusekiDatasetInitializer   <- IOFusekiDatasetInitializer()
+      subscriber                 <- Subscriber(ApplicationLogger)
+      triplesGeneration          <- TriplesGeneration[IO]()
+      sentryInitializer          <- SentryInitializer[IO]()
+      maybeRenkuPythonDevVersion <- RenkuPythonDevVersionConfig[IO]() recoverWith errorToNone
```
joke1196, line 72: replace head of compatibility matrix with dev version if exists

joke1196:
Add insert query for latest cliVersion and schemaVersion to DEV, then PROD.

joke1196:
Bump CliVersion in docker file to the latest

triples-generator/src/main/scala/ch/datascience/triplesgenerator/models/package.scala
travissarles: Travis

joke1196 (approved):

PR #583, opened 2021-01-14
513 el to store status transition time

### LINKED ISSUE #513: Event-log to store statuses transition times
As a renku-graph maintainer, I'd like event-log to store information about how long it took to get an event to a certain status.

**The problem**
Currently, we don't know how long it takes for an event to get to a certain status. For instance, we don't know how long it took to get to the `TRIPLES_STORE` status (from the `PROCESSING` status). Lack of that information prevents us from doing more precise adjustments to the priorities when selecting the next event for processing. As a result, there are situations when we pick-up too many events for big projects so there's not much space left for new events.

**Acceptance criteria:**
- [x] DB initialisation process to create a new table
  - [x] the process to create a new `processing_time` table with columns: `event_id`, `project_id` and `processing_time`;
  - [x] the process to add a compound index on `event_id`, `project_id` (see the `event_log` table initialisation algorithm);
  - [x] the process should do nothing if the table already exists;
- [x] event-log to accept the processing time in the `PATCH /events/:event-id/:project-id` endpoint;
  - [x] the paload to the enpoint to contain an optional `processingTime` property;
  - [x] the `processingTime` property to be contain time unit: `ms`, `s`, `min`, `h`;
  - [x] the enpoint to return `BAD_REQUEST (404)` if `processingTime` has not unit;
  - [x] the enpoint to return `BAD_REQUEST (404)` if `processingTime` < 0;
  - [ ] ~~if there's no `processingTime` in the payload, the logic behind the endpoint should calculate the processing time using the current time and `execution_date` for that event;~~
  - [x] the endpoint to store the processing time in the new `processing_time` table in milliseconds;
  - [x] update event-log's `README.md`;
  - [x] add processingTime to all status change command (to top level process)

joke1196:
closes #513


graph-commons/src/main/scala/ch/datascience/tinytypes/constraints/DurationNotNegative.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package ch.datascience.tinytypes.constraints
+
+import ch.datascience.tinytypes.Constraints
+
+import java.time.Duration
+
+trait DurationNotNegative extends Constraints[Duration] {
+  addConstraint(
+    check = _.toMillis > 0,
```
joke1196, line 27: allow 0 millis value

graph-commons/src/test/scala/ch/datascience/generators/Generators.scala
```diff
@@ -237,6 +236,13 @@ object Generators {
       .map(LocalDateTime.ofInstant(_, ZoneOffset.UTC))
       .map(_.toLocalDate)
 
+  val positiveJavaDurations = javaDurations()
+
+  def javaDurations(min: Long = 1, max: Long = 20000): Gen[Duration] =
```
joke1196, line 241: min default could be 0

jachro (approved):

PR #585, opened 2021-01-19
541 subscription category for generated triples

### LINKED ISSUE #541: EL to allow subscriptions for TRIPLES_GENERATED
As a renku-graph maintainer, I'd like EL to allow subscriptions for `TRIPLES_GENERATED`.

**The problem**
Currently, we've got a single triples generation process which runs `renku log`, transforms the triples and pushes them to the TS. We need to split that so we could be able to use renku-core for `renku log` operation. That would require EL to allow mentioned subscription.

**Acceptance criteria:**
- [x] create a new subscription category which would be delivering all events in the `TRIPLES_GENERATED` status;
- [x] the subscription payload should look like:
  ```json
  {
    "subscriberUrl": "HTTP://host",
    "statuses": ["TRIPLES_GENERATED"]
  }
- [x] add metrics for events waiting for transformation
  ```

joke1196:
close #541


event-log/src/test/scala/io/renku/eventlog/DbEventLogGenerators.scala
```diff
@@ -36,7 +36,6 @@ object DbEventLogGenerators {
   implicit val executionDates: Gen[ExecutionDate] = timestamps map ExecutionDate.apply
```
joke1196, line 36: Merge with Generators and a find a better name

jachro (approved):

PR #597, opened 2021-01-27
chore: add acceptance test action

rokroskar:
Adds the possibility to run renku acceptance tests on PRs. 

/deploy


.github/workflows/acceptance-tests.yaml
```diff
@@ -0,0 +1,113 @@
+name: Acceptance tests
+
+on:
+  pull_request:
+    types:
+    - opened
+    - edited
+    - synchronize
+    - reopened
+    - closed
```
jachro, line 10: Sorry for the stupid question but would that mean that the job will get triggered on all the specified PR statuses? I'm wondering if it wouldn't be better that we make the new acceptance tests step mandatory for our release step and so the new step would need tag-check, test, acceptance-test and formatter to succeed before it gets triggered.

jachro: And I'd say it could be maybe triggered only after merging to master rather than on PR branches.

rokroskar: yes it's triggered on all, but if `/deploy` is missing from the PR description it won't actually deploy and test anything. We could add an additional requirement that it runs on all tags?

jachro: Yes, that might be nice to trigger it on every tag. Actually, maybe only on tags.

rokroskar: right now it's optional to run the tests on PRs - the idea is to catch potential breaking changes early. The PR reviewer can request the PR to have the acceptance tests run or not, depending on the nature of the PR changes. This is the intent anyway, we'll see how well that works.

rokroskar: but I can certainly change this behavior if you don't think it applies to this repo

jachro: Yes, that makes sense. I forgot about the `/deploy` keyword (and sorry, didn't read the change carefully enough). In such a case, it's definitely nice to keep the possibility of running the job even on a PR.

rokroskar: Actually, after thinking about it a bit I think doing it also on tags is a bit more complicated and this action would need to be refactored quite a bit. Furthermore, since tags get pushed to the main renku repo anyway, the acceptance tests will run there regardless. If you'd like to have it run on tags in this repo, I suggest we take that up as a separate issue since there are other concerns (when to do a teardown, for example). What do you think?

jachro: Yes, I think it should be fine to use the action as it is now. As far as I understand, the tests are triggered in an _on-demand_ fashion so that's actually all we need. As you mentioned, the action is kicked off on each tag anyway and that's what fits our development process the most.

jachro (approved):

jachro (approved):

Issue #629, opened 2021-02-22
Zombie events to be chased quicker


jachro:
**Acceptance criteria:**
- [x] need to store info about when an event was sent (a new `event_delivery` table with `event_id`, `project_id`, `delivery_url`);
- [x] need to store info about currently active subscribers (a new `subscriber` table with `delivery_url` and `source_url`);
- [x] the event processes:
  - [x] EL starts and a subscriber subscribes to a category name then we add a row to the `subscriber` table;
  - [x] when EL sends an event we:
    - [x] add a row to the `event_delivery` table if the event got `Delivered` (expect the category is of the broadcast type then we don't add a row);
    - [x] do not add a row if the event got `ServiceBusy`;
    - [x] do not add a row if the event got `Misdelivered` and remove the subscriber from the `subscriber` table;
  - [x] when an event status is changed, a relevant row from the `event_delivery` should be removed (part of the event status change endpoint);
- [x] the zombie event finder:
  - [x] set max processing with a greater duration
  - [x] to check if there are events in `event_delivery` table for subscribers which are not listed in the `subscriber` table; if yes, these events are zombies;
  - [x] zombie finder would need to go through all the records except the one with `source_url` being his urls from the `subscriber` table and:
    - [x] ping the `source_url` and if it's down then remove all the rows for that `source_url` from the `subscriber` table;
  - [x] zombie finder to check if there are events with the `Zombie event` message for longer than 5 mins (use `execution_date` here);
- [x] verify if a service gets a new IP after being restarted; if not then we'd need to create some additional points here about adding some identifier to the subscriber;
  - [x] the service will keep the IP if it gets restarted due to some internal problem e.g. memory issue;
  - [x] each subscriber to send a unique identifier (different each time a subscriber starts-up); the format `yyyymmddhhmmss-{random 4 digits int}`;
  - [x] on adding a subscriber to the registry, its identifier would need to be stored in the `subscriber` table;
  - [x] the `event_delivery` table would have to get a foreign key on the subscriber identifier and the `delivery_url` would have to be dropped;
  - [x] when a subscriber with an IP already existing in the `subscriber` table is trying to subscribe, the old row for that IP would have to be removed;
  - [x] relevant zombie events finders are updated with the changes mentioned above.

PR #632, opened 2021-02-23
629 zombie event quick chase (DB creation)

joke1196:
This PR is part of #629. These first steps allow us to split the work for the next PRs


joke1196:
moved to other branch

PR #633, opened 2021-03-01
629 zombie event quick chase el

### LINKED ISSUE #629: Zombie events to be chased quicker

joke1196:
closes #629

### DIFF
--- a/event-log/README.md
+++ b/event-log/README.md
@@ -522,6 +522,17 @@ Event-log uses relational database as an internal storage. The DB has the follow
 | status          VARCHAR    PK    NOT NULL |
 | processing_time INTERVAL         NOT NULL |
 
+| subscriber                         |
+|------------------------------------|
+| source_url   VARCHAR   PK NOT NULL |
+| delivery_url VARCHAR   PK NOT NULL |
+
+| event_delivery                        |
+|---------------------------------------|
+| event_id     VARCHAR   PK FK NOT NULL |
+| project_id   INT4      PK FK NOT NULL |
+| delivery_url VARCHAR   PK NOT NULL    |
+
 ## Trying out
 
 The event-log is a part of multi-module sbt project thus it has to be built from the root level.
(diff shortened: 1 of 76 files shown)


jachro:
Closing in favour of #634

PR #648, opened 2021-03-15
charts: triples generator to accept number of replicas

pameladelgado:
Noticed that this was missing on the TG charts, useful for Renkulab.
@jachro we might want to update the Chart version but I leave that to you.

/deploy


joke1196:
Merged into https://github.com/SwissDataScienceCenter/renku-graph/pull/647

PR #651, opened 2021-03-16
454 commits synchronization

### LINKED ISSUE #454: Commits synchronisation events to be initiated from the event-log for single projects
As a renku-graph maintainer, I'd like the commits synchronisation process to be based on the subscription model. The commits synchronization process is a safety net in case we deploy and the webhooks don't work. 

### Curent design
- There is a batch job on Webhook which checks now and then (about every hour) if there are commits to be synchronized. There is a major problem: scalability. We can't put more than one instance. Each of the new instances will be doing synchronizing by itself. On Prod, this would kill GitLab. Even with one instance when `WebHook` service polls GitLab, it creates too much load.

**Acceptance-criteria:**
- [x] a new `COMMITS_SYNC` subscription category to be created on the EL;
  [x] the subscription payload should look like this:
  ```json
  {
    "categoryName": "COMMITS_SYNC",
    "subscriber": {
      "url":      "http://host/path",
      "id":       "20210302140653-8641"
    }
  }
  ```
 -  [x]  a sync event should be issued for a single project;
-  [x] a sync event should be issued every:
   - [x] for projects with `latest_event_date` <= 7 days -> hour
   - [x] for projects with `latest_event_date` > 7 days -> day
 - [x] the category should update a relevant row in the `subscription_category_sync_time` on issuing an event;
-  [x] update README;

- [x] a new XXX service to be created with the logic copied from the relevant process in the webhook-service
  - [x]  the name of the service should be good enough to describe the purpose of it (commits synchronisation, detect and create new events for force push cases, project renames, project deletions, new commit events creation from a push event etc)
  - [x] the subscription should be built using the relevant functionality for the graph commons;
  - [x] the service to expose a `POST /events` endpoint for accepting the `COMMITS_SYNC` events;
  - [x] the event from EL to Commit Event Service should look like follows:
  ```json
  {
    "categoryName": "COMMITS_SYNC",
    "id": "df654c3b1bd105a29d658f78f6380a842feac879",
    "project": {
      "id":   123,
    },
    "lastSynced": "2001-09-04T10:48:29.457Z"
  }
  ```
  - [x]  for a single project, compare the two commit Ids: one from EL and the other from GitLab to check latest commit. 
    - [x] if there is a difference, sync events for that project and post them to event log
  - [x] the endpoint should have the same logic as the `POST /events` in EL or TG; 
    - [ ] it should call the toEvent method like in EL
  - [x] the service should subscribe to the `COMMITS_SYNC` category
  - [x] the service should be able to process `COMMITS_SYNC` events sent by EL (the logic can be found in the webhook-service, see `MissedEventsLoader#loadEvents`)
  - [x] create README;
  - [x] create helm chart and make the new service to be built during charpress run;
  - [x] clean-up webhook-service;
  - [ ] add panels showing JVM metrics for the new service;
- [x] add acceptance test (similar to `ZombieEventDetectionSpec`);

- [x] the `GET /events?latest_per_project=true` on the EL should be removed;
  - update README;

travissarles:
closes #454 
/deploy


acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/CommitSyncFlowsSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+        sql"""|DELETE FROM subscription_category_sync_time 
+              |WHERE project_id = ${projectId.value} AND category_name = 'COMMIT_SYNC'
+              |""".stripMargin.update.run
+      }
+      sleep((4 seconds).toMillis)
+
+      And("commit events for the missed event are created and processed")
+      `wait for events to be processed`(project.id)
+
+      Then("triples for the missed commit should be in the RDF Store")
```
travissarles, line 119: Perhaps we should also check that if another Push Event arrives, that nothing happens

acceptance-tests/src/test/scala/ch/datascience/graph/acceptancetests/stubs/GitLab.scala
```diff
@@ -119,11 +102,36 @@ object GitLab {
...
+          }                         
+        ]""".noSpaces))
+      }
+    else {
+      val getLatestCommitWithScenario = getLatestCommit.inScenario(s"fetch latest commit for $projectId")
+
+      commitIds.zipWithIndex foreach { case (commitId, idx) =>
+        stubFor {
+          getLatestCommitWithScenario
+            .whenScenarioStateIs(if (idx == 0) Scenario.STARTED else s"call $idx")
```
travissarles, line 132: Perhaps you can explain this one during the showcase :)

jachro (approved):

PR #706, opened 2021-06-30
chore: moving to renku-actions

jachro:
/deploy


travissarles (approved):

ableuler:
I would recommend you to pin the action version by replacing `@master` with `@v0.1.0`, so a faulty change to the `renku-actions` repo won't break you here.

jachro:
@ableuler we'll do that.

Issue #660, opened 2021-03-29
Change the re-provisioning process


jachro:
Prerequisites:
- [x] try out the new renku graph generation functionality:
  - [x] how does this work for repos having commit(s) that fails generation, e.g. there are 3 commits on a repo, the first and the last one are fine for graph generation but the 2nd fails; when we do graph generation after the 3rd commit, we should get graph for the commits before the 2nd one (the failing one); this is to prove that running `generate graph` on all the commits produces the same result as running it on the very last commit only;
  - [x] to work out the sequence of steps to use the new graph generation in TG, e.g. Renku migrate -> check if the graph generation was already done on the repo (to look at some JSON file somewhere in the .renku) -> Renku graph generate
  
  * I took the new `renku graph` functionality for a test drive on few projects. It worked well on small projects as well as on big ones like `covid-19-public-data`. The order of operations TG to follow should be like this:
    * `renku migrate`
    * `renku graph generate`; if the process exists with `0` or `1` and the `Error: Graph metadata exists. Use ``--force`` to regenerate it.` message, all is good; otherwise, things went wrong;
    * `renku graph export` returns the JSON-LD;

**Acceptance criteria:**
- [x] keeping in mind that the new graph generation does not include info about files that are not involved in datasets or workflows, clean-up code which generates such objects (maybe smth in the bundles?);
- [x] make sure the `TRIPLES_GENERATED` category finder does issue events only for the latest events in status `TRIPLES_GENERATED` which have a payload;
- [x] add a new `EVENTS_STATUS_CHANGE` consume event category which should:
  - [x] be changing statuses for the given event as well as all the events older than the given one to `TRIPLES_GENERATED`; events in status `SKIPPED` should not be touched;
  - [x] event could be changing status (depending on the data in the event) for: ~all the events (to `NEW`), for all the project events (to `NEW`),~ for all the project events older than given event id (to `TRIPLES_GENERATED` or `TRIPLES_STORE`)
  - [x] When changing status to `NEW` remove payload and processing time
  - [x] `AWAITING_DELETION` event should be removed 
  - [x] should we migrate the _all events to `NEW`_ status change to the new event type?
    - [x] switch TG to the new event
    - [x] clean-up the old endpoint doing the same task
- [x] change the steps in the triples generation flow to accommodate the new Renku graph generation process (see prerequisites);
~- [ ] add a new row to the `compatibility-matrix` to kick-off re-provisioning process on the next deployment;~
[...]

PR #723, opened 2021-08-04
refactor: new provisioning process (#660)

travissarles:
to be closed


jachro:
We'll create another PR

PR #736, opened 2021-08-20
feat: 680 awaiting deletion event log

### LINKED ISSUE #680: CES service to mark events as AWAITING_DELETION
**Acceptance criteria:**
* PR 1
  - [x] add a new AWAITING_DELETION status to EL;
  - [x] EL would have to be able to work with AWAITING_DELETION status;
  - [x] COMMIT_SYNC event from EL arrives at CES;
  - [x] CES collect all the commit ids created between now and the commit id from the event;
  - [x] CES creates new events for all the commits EL does not know about;
  - [x] CES changes the status of commits in EL that does not exist in GitLab;
  - [x] if both the commit id and the commit count matches with the findings from GitLab, CES can assume nothing has to be done
  - [x] if at least one for the pieces of data do not match, then the full list of commit ids has to be pulled from GitLab and EL so CES can reason about which commits need to be created or deleted (CES will create NEW/SKIPPED/AWAITING_DELETION events on EL)
  - [x] the handler on the CES side should throttle the events and process one event at a time (similarly as triples generation on the TG);
* PR 2
  - [x] there's a new GLOBAL_COMMIT_SYNC category that has its own schedule;
  - [x] the event should contain info about the latest commit id and the number of commits for a project
    - event part:
     ```json
     {
      "categoryName": "GLOBAL_COMMIT_SYNC",
      "project": {
        "id": 1213,
        "path": "namespace/name"
      }
     }
     ```
     - payload part:
     ```json
     {
      "eventsCount": 123,
      "latestCommitId": "sha",
      "commitIds": ["sha", "sha1", ..., "shax"]
     }
     ```

  - [x] This process would happen less than once per day because it's very costly (perhaps once a week). This needs to happen because sometimes the commit sync category process breaks;
  - [x] Implement an event finder on EL which would create new events every week. See `MemberSyncEventsFinder` as an example
  - [x] When fetching global commit sync event and sending it to the CES the return status could be Busy. In this case the event last synced date should be reverted to it's original date
- [x] implement `show` for GlobalCommitSync event types on EL and CES

in GitLab 1 2 3 4 5 -6 -7 8
in EL: 1 3 5 6 7
CES

the order of steps:
* COMMIT_SYNC comes with 7
* CES fetches the latest commit from GL -> 8
* CES checks if 8 exists in EL -> get NO (if it would be YES then nothing to continue)
* CES sends CREATION for 8
* CES fetches the parent of 8 -> 5
* CES checks if 5 exists in EL -> gets YES

deletion subprocess
* CES checks if 7 exists in GL -> NO
* CES sends AWAITING_DELETION to EL for 7
* CES gets event info (parent) of 7 from EL -> 6
* CES checks if 6 exists in GL -> NO
* CES sends AWAITING_DELETION to EL for 6
* CES gets event info (parent) of 6 from EL -> 5
* CES checks if 5 exists in GL -> YES

Another thing to think about:
* COMMIT_SYNC comes with 7
* CES gets 404 when fetching the latest commit from GL
* CES continues as in the `deletion subprocess`

7
fetch page of 5 from GL 2 3 4 5 8
check in GL if 7 exists:
if yes get previous pages until we get 7
if no tell EL that 7 is for deletion
send COMMIT_SYNC_REQUEST to EL and we start from the beginning (from 6 this time)

joke1196:
closes #680 
/deploy


jachro (approved):
Great job! Thanks David.

PR #786, opened 2021-11-22
feat: allow command to be optional

### LINKED ISSUE #781: Handle blank Command name when decoding the model
When decoding a project it is possible to encounter a blank command 
```
{
    "@id": "https://renkulab.io/plans/1acd4b32-4aee-44ba-8afc-c05dba3ccdb9",
    "@type": [
      "http://schema.org/Action",
      "http://schema.org/CreativeWork",
      "http://www.w3.org/ns/prov#Plan"
    ],

    "https://swissdatasciencecenter.github.io/renku-ontology#command": [
      {
        "@value": ""
      }
    ],
...
}
```
At the moment this will cause the decoding to fail but it is a valid state and should be handled acordingly. This is how it can happen: 
```
bash myscript.sh : bash=command, myscript.sh = first input
/bin/bash myscript.sh : /bin/bash=command (file exists but is not in project), myscript.sh=first input
myscript.sh (with shebang in script): command=nothing, myscript.sh =first input (if myscript.sh is in the project)
```

travissarles:
closes #781

### DIFF
--- a/commit-event-service/src/main/scala/io/renku/commiteventservice/events/categories/globalcommitsync/eventgeneration/ELCommitFetcher.scala
+++ b/commit-event-service/src/main/scala/io/renku/commiteventservice/events/categories/globalcommitsync/eventgeneration/ELCommitFetcher.scala
@@ -1,3 +1,21 @@
+/*
+ * Copyright 2021 Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.renku.commiteventservice.events.categories.globalcommitsync.eventgeneration
 
 import cats.MonadThrow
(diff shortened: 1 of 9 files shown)


renku-model/src/main/scala/io/renku/graph/model/entities/Plan.scala
```diff
@@ -79,7 +79,7 @@ object Plan {
       resourceId            <- cursor.downEntityId.as[ResourceId]
       name                  <- cursor.downField(schema / "name").as[Name]
       maybeDescription      <- cursor.downField(schema / "description").as[Option[Description]]
-      command               <- cursor.downField(renku / "command").as[Command]
+      command               <- cursor.downField(renku / "command").as[Option[Command]]
```
travissarles, line 82: rename to maybeCommand

commit-event-service/src/main/scala/io/renku/commiteventservice/events/categories/globalcommitsync/eventgeneration/ELCommitFetcher.scala
```diff
@@ -1,3 +1,21 @@
+/*
+ * Copyright 2021 Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
```
travissarles, line 9: revert this header

renku-model/src/test/scala/io/renku/graph/model/testentities/generators/ActivityGenerators.scala
```diff
@@ -53,6 +54,8 @@ trait ActivityGenerators {
   implicit val planDescriptions: Gen[plans.Description] =
     sentences().map(_.value).toGeneratorOf[plans.Description]
   implicit val planCommands: Gen[plans.Command] = nonBlankStrings().map(_.value).toGeneratorOf[plans.Command]
+  implicit val planCommandsOptions: Gen[Option[plans.Command]] =
```
travissarles, line 57: Check if we need this or if it can be derived

PR #803, opened 2021-12-01
feat: Metrics for GitLab api calls

### LINKED ISSUE #796: GitLab API call metrics
### Current situation
We don't have visibility into the number of calls to each endpoint. At the moment, the throttler will prevent the services from making too many calls to GitLab and crashing it. However, this is a fixed 100 calls per second. Knowing more about how the GitLab API is called by our services, we might be able to find smarter ways to throttle calls.

### Proposed solution
Create a GitLab Client in the code, similar to the RdfStoreClient which will record metrics about the number of calls to each endpoint. 

### Acceptance criteria

- [x]  Use Grafana explorer to see if we could use the execution times from the Sparql queries to see the rate, the number of calls 
- [ ] Create a GitLab client which extends RestClient and includes a histogram gauge
  - [ ] Add it as a dependency to all clients which use GitLab such as GitLabCommitFetcher
  - [x] Move `timeRecorder`, `retryInterval`, `maxretries`, `requestTimeoutOverride` to client
  - [x] Based on the needs of each dependent class, implement the minimal API to call the `RestClient` methods. Use `RestClient.secureRequest` as an example.
  - [x] `validateUri` should be done in the `GitLabClient` 
- [x] When collecting metrics, keep the naming conventions from the GitLab API 
- [ ] Record global totals for each endpoint

travissarles:
closes #796 

/deploy


commit-event-service/src/main/scala/io/renku/commiteventservice/Microservice.scala
```diff
@@ -44,12 +46,16 @@ object Microservice extends IOMicroservice {
     certificateLoader     <- CertificateLoader[IO]
     sentryInitializer     <- SentryInitializer[IO]
     gitLabRateLimit       <- RateLimit.fromConfig[IO, GitLab]("services.gitlab.rate-limit")
+    gitLabUrl             <- GitLabUrlLoader[IO]()
```
jachro, line 49: Could we maybe move gitLabUrl and apiCallRecorder to the apply of GitLabClient?

commit-event-service/src/test/scala/io/renku/commiteventservice/events/categories/globalcommitsync/eventgeneration/gitlab/GitLabCommitFetcherSpec.scala
```diff
@@ -20,156 +20,250 @@ package io.renku.commiteventservice.events.categories.globalcommitsync.eventgene
...
 
       gitLabCommitFetcher
-        .fetchGitLabCommits(projectId, pageRequest)(maybeAccessToken = None)
-        .unsafeRunSync() shouldBe PageResult(commits = Nil, maybeNextPage = None)
+        .fetchGitLabCommits(projectId, pageRequest)(maybeAccessToken)
+        .unsafeRunSync() shouldBe expectation
     }
 
-    "return an empty list if project for NOT_FOUND" in new TestCase {
+    /////////////////// ResponseMapper tests
```
jachro, line 121: Do you find this comment helpful?

commit-event-service/src/test/scala/io/renku/commiteventservice/events/categories/globalcommitsync/eventgeneration/gitlab/GitLabCommitFetcherSpec.scala
```diff
@@ -20,156 +20,250 @@ package io.renku.commiteventservice.events.categories.globalcommitsync.eventgene
...
+          )
           .unsafeRunSync()
-      }.getMessage should startWith(
-        s"GET $gitLabApiUrl/projects/$projectId/repository/commits?page=${pageRequest.page}&per_page=${pageRequest.perPage} returned ${Status.Ok}; error: Invalid message body: Could not decode JSON: {}"
-      )
+      }
     }
   }
 
+  ///////////// Single commit tesst
```
jachro, line 182: Same as above and there's one more below

graph-commons/src/main/scala/io/renku/http/client/GitLabClient.scala
```diff
@@ -0,0 +1,79 @@
...
+    )
+    with GitLabClient[F] {
+
+  def send[ResultType](method: Method, path: Uri, endpointName: String Refined NonEmpty)(
+      mapResponse:             ResponseMapping[ResultType]
+  )(implicit
+      maybeAccessToken: Option[AccessToken]
+  ): F[ResultType] = for {
+    uri     <- validateUri(show"$gitLabApiUrl/$path")
+    request <- request(method, uri, maybeAccessToken, endpointName)
```
jachro, line 49: We could maybe use the secureRequest instead of request?

graph-commons/src/test/scala/io/renku/http/client/GitLabClientSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+
+      client.send(method, path, endpointName)(_ => IO(result)).unsafeRunSync() shouldBe result
+    }
+
+    "fetch commits using the given personal access token" in new TestCase {
+      val personalAccessToken = personalAccessTokens.generateOne
+
+      stubFor {
+        callMethod(method, s"api/v4/$path")
+          .withHeader("PRIVATE-TOKEN", equalTo(personalAccessToken.value))
```
jachro, line 53: We've got this `withAccessToken` extension in the ExternalServiceStubbing to test token existence so maybe we could use it here for this test so the one for oauth token can go.

commit-event-service/src/test/scala/io/renku/commiteventservice/events/categories/globalcommitsync/eventgeneration/gitlab/GitLabCommitFetcherSpec.scala
travissarles: It was helpful during development but I can remove it now.

graph-commons/src/test/scala/io/renku/http/client/GitLabClientSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+          .willReturn(
+            okJson(result)
+              .withHeader("X-Next-Page", maybeNextPage.map(_.show).getOrElse(""))
+          )
+      }
+
+      client.send(method, path, endpointName)(_ => IO(result)).unsafeRunSync() shouldBe result
+    }
+
+    "Handle empty JSON" in new TestCase {
```
jachro, line 78: I'm afraid this test and the subsequent three are not achieving what they suppose to. The reason for that is the `mappingResponse` function that is passed to the send method does not use/interact with the response. I believe the idea here is to give some mapping function that tries, for instance in this particular test, to decode some value from the response (in other cases fail if the response's status is different than the status in the mapping function, and so on).

graph-commons/src/test/scala/io/renku/http/client/GitLabClientSpec.scala
```diff
@@ -0,0 +1,191 @@
...
+            okJson(result)
+              .withHeader("X-Next-Page", maybeNextPage.map(_.show).getOrElse(""))
+          )
+      }
+
+      client.send(method, path, endpointName)(_ => IO(result)).unsafeRunSync() shouldBe result
+    }
+
+    "fetch commits using the given personal access token" in new TestCase {
+      val personalAccessToken = personalAccessTokens.generateOne
```
jachro, line 51: What I was thinking here is that we could generate some access token, not specifically personalAccessToken, use the `withAccessToken` as you did and simply delete the `fetch commits using the given oauth token` spec as all will be tested in this one.

graph-commons/src/test/scala/io/renku/http/client/GitLabClientSpec.scala
```diff
@@ -0,0 +1,191 @@
...
+          )
+      }
+
+      private implicit lazy val mapResponse: PartialFunction[(Status, Request[IO], Response[IO]), IO[Option[Int]]] = {
+        case (Ok, _, response)    => response.as[List[Int]].map(_.headOption)
+        case (NotFound, _, _)     => Option.empty[Int].pure[IO]
+        case (Unauthorized, _, _) => UnauthorizedException.raiseError[IO, Option[Int]]
+      }
+
+      intercept[Exception](client.send(method, path, endpointName)(mapResponse).unsafeRunSync())
```
jachro, line 139: I guess we could maybe try to assert if the intercepted exception is the `UnauthorizedException`

graph-commons/src/test/scala/io/renku/http/client/GitLabClientSpec.scala
```diff
@@ -0,0 +1,191 @@
...
+
+    "Handle 404 NOT FOUND without throwing" in new TestCase {
+      stubFor {
+        callMethod(method, s"api/v4/$path")
+          .willReturn(
+            notFound()
+          )
+      }
+
+      client.send(method, path, endpointName)(_ => IO(())).unsafeRunSync() shouldBe ()
```
jachro, line 102: What if we make the mapping from lines 114, 133 and 149 global and use it in all the tests?

graph-commons/src/test/scala/io/renku/http/client/GitLabClientSpec.scala
```diff
@@ -0,0 +1,162 @@
...
+
+      stubFor {
+        callMethod(method, s"api/v4/$path")
+          .willReturn(
+            okJson(result)
+              .withHeader("X-Next-Page", maybeNextPage.map(_.show).getOrElse(""))
+          )
+      }
+
+      client.send(method, path, endpointName)(_ => IO(result)).unsafeRunSync() shouldBe result
```
jachro, line 47: Travis, why not use the `mappingResponse` from the TestCase all the tests here where we do `(_ => IO(result))`. I think it would make the tests more robust.

jachro (approved):
Thanks Travis

PR #816, opened 2022-01-11
chore: 805 docker init process

### LINKED ISSUE #805: Use proper init process as entrypoint
The KG services often spawn long-running child processes (e.g. `renku graph export`) (in particular the triples generator if I understand things correctly). This can lead to zombie process and generally make it hard for k8s to properly terminate containers. Using a proper init process as entrypoint should solve this problem, see for example this blog post: https://ahmet.im/blog/minimal-init-process-for-containers/

I think that using [tini](https://github.com/krallin/tini) should be good enough here, in which case this might be a very simple fix. The acceptance criterium for this issue is that termination signals sent to the container are properly forwarded to long running renku or git processes.

joke1196:
closes #805 
/deploy renku=1.0-next renku-ui=1.0-next renku-core=develop

### DIFF
--- a/commit-event-service/Dockerfile
+++ b/commit-event-service/Dockerfile
@@ -25,8 +25,8 @@ COPY --from=builder /work/commit-event-service/target/universal/stage .
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 ENV TZ UTC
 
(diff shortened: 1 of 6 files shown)


jachro (approved):
👍

PR #841, opened 2022-02-02
fix: event changing project events to NEW to remove AWAITING_DELETION

jachro:
This PR tweaks the `STATUS_CHANGE_EVENT` changing all project's events to `NEW` so it removes all `AWAITING_GENERATION` events belonging to it.

### DIFF
--- a/event-log/src/main/scala/io/renku/eventlog/events/categories/statuschange/ProjectEventsToNewUpdater.scala
+++ b/event-log/src/main/scala/io/renku/eventlog/events/categories/statuschange/ProjectEventsToNewUpdater.scala
@@ -32,31 +32,34 @@ import io.renku.eventlog.{EventDate, ExecutionDate}
 import io.renku.events.consumers.Project
 import io.renku.graph.model.events.EventStatus
 import io.renku.graph.model.projects
-import io.renku.metrics.LabeledHistogram
+import io.renku.metrics.{LabeledGauge, LabeledHistogram}
 import org.typelevel.log4cats.Logger
+import skunk.data.Completion
 import skunk.implicits._
 import skunk.{Session, ~}
 
 import java.time.Instant
 import scala.util.control.NonFatal
-import skunk.data.Completion
-import io.renku.metrics.LabeledGauge
 
 private trait ProjectEventsToNewUpdater[F[_]] extends DBUpdater[F, ProjectEventsToNew]
 
 private object ProjectEventsToNewUpdater {
   def apply[F[_]: Async: Logger](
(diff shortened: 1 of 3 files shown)


event-log/src/main/scala/io/renku/eventlog/events/categories/statuschange/ProjectEventsToNewUpdater.scala
```diff
@@ -131,23 +133,26 @@ private class ProjectEventsToNewUpdaterImpl[F[_]: Async: Logger](
...
       .mapResult {
         case Completion.Delete(count) => count
         case _                        => 0
       }
   }
 
-  private def cleanUpDeletingMetrics(project: Project, removeProjectDeletingEvents: Int) =
-    Kleisli.liftF(deletingEventGauge.update((project.path, -removeProjectDeletingEvents.toDouble)))
+  private def cleanUpAwaitingDeletionMetrics(project: Project, eventsCount: Int) =
+    deletingEventGauge.update((project.path, -eventsCount.toDouble))
```
joke1196, line 152: I think there is a gauge for the awaiting deletion event as well. Not sure if we should combine these two.

event-log/src/test/scala/io/renku/eventlog/events/categories/statuschange/ProjectEventsToNewUpdaterSpec.scala
```diff
@@ -190,16 +188,16 @@ class ProjectEventsToNewUpdaterSpec
...
-        case EventStatus.AwaitingDeletion                            => zippedEventPayloads.generateOption
+        case AwaitingDeletion                                        => zippedEventPayloads.generateOption
         case _                                                       => zippedEventPayloads.generateNone
       },
-      projectPath = projectPath
+      projectPath = project.path
     )
 
     status match {
       case EventStatus.TriplesGenerated | EventStatus.TriplesStore =>
```
joke1196, line 198: I guess we should remove as well the `EventStatus` from the first case

event-log/src/test/scala/io/renku/eventlog/events/categories/statuschange/ProjectEventsToNewUpdaterSpec.scala
```diff
@@ -190,16 +188,16 @@ class ProjectEventsToNewUpdaterSpec
       },
       maybeEventPayload = status match {
         case EventStatus.TriplesStore | EventStatus.TriplesGenerated => zippedEventPayloads.generateSome
```
joke1196, line 190: Same as line 198

joke1196 (changes requested):

PR #843, opened 2022-02-03
fix: commit sync processes to work with invalid access token

jachro:
It looks there are cases when we end up with some invalid access tokens for a project that is deleted. It looks it's safe for the sync processes to treat such cases as if the project is removed.

/deploy

### DIFF
--- a/commit-event-service/src/main/scala/io/renku/commiteventservice/events/categories/commitsync/eventgeneration/LatestCommitFinder.scala
+++ b/commit-event-service/src/main/scala/io/renku/commiteventservice/events/categories/commitsync/eventgeneration/LatestCommitFinder.scala
@@ -18,9 +18,7 @@
 
 package io.renku.commiteventservice.events.categories.commitsync.eventgeneration
 
-import cats.data.OptionT
 import cats.effect.Async
-import cats.effect.kernel.Temporal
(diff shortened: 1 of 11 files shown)


commit-event-service/src/main/scala/io/renku/commiteventservice/events/categories/commitsync/eventgeneration/LatestCommitFinder.scala
```diff
@@ -78,9 +69,7 @@ private class LatestCommitFinderImpl[F[_]: Async: Temporal: Logger](
 }
 
 private object LatestCommitFinder {
-  def apply[F[_]: Async: Temporal: Logger](
-      gitLabThrottler: Throttler[F, GitLab]
-  ): F[LatestCommitFinder[F]] = for {
+  def apply[F[_]: Async: Temporal: Logger](gitLabThrottler: Throttler[F, GitLab]): F[LatestCommitFinder[F]] = for {
```
joke1196, line 72: We could remove Temporal as Async extends it

travissarles (approved):

joke1196 (changes requested):

joke1196 (approved):

PR #846, opened 2022-02-04
build: update sentry values

lorenzo-cavazzi:
Align sentry values with the other renku repositories.

Reference: https://swiss-data-science.slack.com/archives/G67N59QL8/p1643901194490749

/deploy


jachro (approved):
Thanks Lorenzo

PR #856, opened 2022-02-09
feat: global commit sync trigger after commit sync 

### LINKED ISSUE #833: CommitSync process should trigger a GlobalCommitSync
After the Commit Sync process is done it should send a event to trigger the GlobalCommit Sync. The Global commit sync will only happen if the number of commits is different and if the latest commit is different. This means it should in general do nothing after the CommitSync process. But in the case of a force push to master then we would get a proper clean up of the commits.

joke1196:
feat: added global commit sync request handling
closes #833 

/deploy #persist


jachro (approved):
Thanks David!

PR #886, opened 2022-02-28
fix: added retry mechanism when storing and fetching the access token

joke1196:
fix: added retry mechanism when storing the access token

### DIFF
--- a/token-repository/src/main/scala/io/renku/tokenrepository/repository/association/TokenAssociator.scala
+++ b/token-repository/src/main/scala/io/renku/tokenrepository/repository/association/TokenAssociator.scala
@@ -20,7 +20,6 @@ package io.renku.tokenrepository.repository.association
 
 import cats.MonadThrow
 import cats.effect.Async
-import cats.effect.kernel.Temporal
 import cats.syntax.all._
 import io.renku.db.{SessionResource, SqlStatement}
 import io.renku.graph.model.projects.{Id, Path}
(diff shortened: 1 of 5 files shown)


token-repository/src/main/scala/io/renku/tokenrepository/repository/association/TokenAssociator.scala
```diff
@@ -53,18 +58,49 @@ private class TokenAssociatorImpl[F[_]: MonadThrow](
...
+  ): F[Unit] = for {
+    _ <- persistAssociation(projectId, projectPath, encryptedToken)
+    _ <- verifyTokenIntegrity(projectPath) recoverWith retry(projectId, projectPath, encryptedToken, numberOfRetries)
+  } yield ()
+
+  private def verifyTokenIntegrity(projectPath: Path) =
+    tokenFinder.findToken(projectPath).value.flatMap {
+      case Some(savedToken) => decrypt(savedToken).void
+      case _ =>
+        new Exception(show"Token associator - saved encrypted token cannot be found for project: $projectPath")
```
jachro, line 77: Should we maybe raise the exception only when we run out of attempts?

joke1196: Yes good idea!

jachro (approved):

PR #938, opened 2022-03-18
refactor: COMMIT_SYNC_REQUEST to be sent with EventSender

jachro:
The `COMMIT_SYNC_REQUEST` events were probably the last ones that were not going through the `EventSender` and so the `sent_events_count` metric was not collecting data for it. This PR fixes this.

### DIFF
--- a/webhook-service/src/main/scala/io/renku/webhookservice/CommitSyncRequestSender.scala
+++ b/webhook-service/src/main/scala/io/renku/webhookservice/CommitSyncRequestSender.scala
@@ -18,70 +18,34 @@
 
 package io.renku.webhookservice
 
-import cats.Eval
+import cats.MonadThrow
 import cats.effect.Async
-import cats.effect.kernel.Temporal
 import cats.syntax.all._
-import eu.timepit.refined.api.Refined
-import eu.timepit.refined.numeric.NonNegative
-import io.renku.control.Throttler
-import io.renku.graph.config.EventLogUrl
-import io.renku.http.client.RestClient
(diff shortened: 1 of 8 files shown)


webhook-service/src/main/scala/io/renku/webhookservice/CommitSyncRequestSender.scala
```diff
@@ -18,70 +18,34 @@
...
 
-  private def waitAndRetry(retry: Eval[F[Unit]], errorMessage: String) = for {
-    _      <- Logger[F].error(s"Sending commit sync request event failed - retrying in $retryDelay - $errorMessage")
-    _      <- Temporal[F] sleep retryDelay
-    result <- retry.value
-  } yield result
+  def sendCommitSyncRequest(syncRequest: CommitSyncRequest, processName: String): F[Unit] =
+    sendEvent(
+      EventRequestContent.NoPayload(syncRequest.asJson),
+      EventSender.EventContext(CategoryName("COMMIT_SYNC_REQUEST"),
```
travissarles, line 45: should we store COMMIT_SYNC_REQUEST in a shared variable somewhere?

travissarles (approved):

webhook-service/src/main/scala/io/renku/webhookservice/CommitSyncRequestSender.scala
jachro: Hm, that might be good and bad as that's the only usage in this service. All the other usages are in EL and it won't be nice to make them a hard link between these two. Anyway, we'd need to think more about how to eventually do it in other places too. For now, it somehow follows the common pattern of not sharing the model classes between different components so they are not tightly coupled.

PR #956, opened 2022-03-25
feat: TG to subscribe for TS_MIGRATION_REQUEST

jachro:
/deploy #persist

### DIFF
--- a/event-log/src/main/scala/io/renku/eventlog/tsMigrationModel.scala
+++ b/event-log/src/main/scala/io/renku/eventlog/tsMigrationModel.scala
@@ -26,9 +26,9 @@ import io.renku.tinytypes.{InstantTinyType, StringTinyType, TinyTypeFactory}
 
 import java.time.Instant
 
-private sealed trait MigrationStatus extends StringTinyType with Product with Serializable
+sealed trait MigrationStatus extends StringTinyType with Product with Serializable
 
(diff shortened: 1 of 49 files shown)


event-log/src/main/scala/io/renku/eventlog/tsMigrationModel.scala
```diff
@@ -16,37 +16,19 @@
...
-
-private object MigratorSubscriptionInfo {
-  implicit lazy val show: Show[MigratorSubscriptionInfo] = Show.show { info =>
-    import info._
-    show"subscriber = $subscriberUrl, id = $subscriberId, version = $subscriberVersion"
-  }
-}
-
-private sealed trait MigrationStatus extends StringTinyType with Product with Serializable
+sealed trait MigrationStatus extends StringTinyType with Product with Serializable
```
travissarles, line 29: Why is this public now?

travissarles (approved):

event-log/src/main/scala/io/renku/eventlog/tsMigrationModel.scala
jachro: It's because it's shared between `TS_MIGRATION_REQUEST` and `MIGRATION_STATUS_CHANGE` categories.

PR #960, opened 2022-03-30
chore(ci): update renku-actions to 0.5.1

olevski:
This should lead to better cleanup of renku CI deployments.


jachro (approved):
Thanks Tasko

PR #977, opened 2022-04-13
feat: enforcing at least one DS creator

jachro:
This PR enforces at least one creator on a Datasets


knowledge-graph/src/main/scala/io/renku/knowledgegraph/datasets/model.scala
```diff
@@ -49,7 +49,7 @@ object model {
                                       sameAs:           SameAs,
                                       versions:         DatasetVersions,
                                       maybeDescription: Option[Description],
-                                      creators:         Set[DatasetCreator],
+                                      creators:         List[DatasetCreator],
```
travissarles, line 52: Is there a reason this isn't a NonEmptyList?

travissarles (approved):

knowledge-graph/src/main/scala/io/renku/knowledgegraph/datasets/model.scala
jachro: It's only due to pragmatic reasons :) If I'd like it to be a NonEmptyList, I'd need to change a bit of the way how the data is collected in the free-text DS search. At the end, this data structure is used only for transferring data between the sparql query and HTTP response, so it's not that bad.

PR #979, opened 2022-04-13
chore: use userid 1000 and k8s security context

### LINKED ISSUE #1000: Run renku-graph as non-root

olevski:
It is great that things were already setup to run as non-root. I just wanted to standardize all the Renku services to use the same user and group id of 1000. I already opened a bunch of PRs with 1000 and it was already used elsewhere. So I will go with 1000 here too.

Also adding the security context is useful because it prevents changes or switching to a higher permissions group or user when the container is running. 

/deploy #persist

closes https://github.com/swissdatasciencecenter/renku-notebooks/issues/1000

### DIFF
--- a/commit-event-service/Dockerfile
+++ b/commit-event-service/Dockerfile
@@ -29,9 +29,9 @@ ENV TZ UTC
 RUN apk add --no-cache tzdata curl bash tini && \
     chown -R daemon:daemon .
 
-ENV GID=90060
+ENV GID=1000
 RUN addgroup -g "$GID" -S cesgroup && \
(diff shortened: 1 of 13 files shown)


olevski:
The way of defining the security context in the values file and injecting it everywhere is exactly what helm does when it creates a brand new helm chart for you with the CLI. So I am just following that pattern.

jachro (approved):

olevski:
By doing stuff manually from the deployment I can launch regular and anonymous sessions without a problem. So the acceptance tests are being flaky or they need an update.

PR #1000, opened 2022-05-26
feat: API docs for lineage endpoint

### LINKED ISSUE #199: Create OpenAPI Json
As a Renku user, I'd like to be able to see graph-services API documentation.

Acceptance criteria:
- [x] Create a model based on the OpenAPI Specification
- [ ] Include security schemes (see below)
- [ ] Serve the JSON at /knowledge-graph/spec.json
- [ ] Add the link to the Json [here](https://github.com/SwissDataScienceCenter/renku/blob/23dde6ba6e24d2ac66685f203f211aef32fa3287/helm-chart/renku/templates/swagger.yaml#L29-L30)

```json
    "securitySchemes": {
      "gitlab-token": {
        "in": "header",
        "name": "Authorization",
        "type": "apiKey"
      },
      "jwt": {
        "in": "header",
        "name": "Renku-User",
        "type": "apiKey"
      },
      "oidc": {
        "openIdConnectUrl": "/auth/realms/Renku/.well-known/openid-configuration",
        "type": "openIdConnect"
      }
    },
  "security": [
    {
      "oidc": []
    },
    {
      "JWT": [],
      "gitlab-token": []
    }
  ]
```

travissarles:
feat: API docs for lineage endpoint

/deploy #persist


knowledge-graph/src/main/scala/io/renku/knowledgegraph/MicroserviceRoutes.scala
```diff
@@ -126,6 +127,7 @@ private class MicroserviceRoutes[F[_]: MonadThrow](
   private lazy val nonAuthorizedRoutes: HttpRoutes[F] = HttpRoutes.of[F] {
     case GET -> Root / "knowledge-graph" / "graphql" => schema()
     case GET -> Root / "ping"                        => Ok("pong")
+    case GET -> Root / "spec.json"                   => docsEndpoint.`get /docs`
```
jachro: Should we maybe choose a name for the API that is more in line with the other APIs? Perhaps, something like `api-spec`?

jachro: Once you decide on the API could you adapt both the name of the endpoint val and the `def` on it, please?

knowledge-graph/src/main/scala/io/renku/knowledgegraph/docs/Endpoint.scala
```diff
@@ -0,0 +1,88 @@
...
+
+private class EndpointImpl[F[_]: Async] extends Http4sDsl[F] with Endpoint[F] {
+  import Encoders._
+
+  override def `get /docs`: F[http4s.Response[F]] = Ok(doc.asJson)
+
+  lazy val doc: OpenApiDocument =
+    OpenApiDocument(
+      "3.0.3",
+      Info("Knowledge Graph API", "Get info about datasets, users, activities, and other entities".some, "1.0.0")
```
jachro: Is the `1.0.0` the version of KG? If yes, we could maybe use the `ServiceVersion.readFromConfig`.

knowledge-graph/src/main/scala/io/renku/knowledgegraph/docs/Endpoint.scala
```diff
@@ -0,0 +1,88 @@
...
+    ).addServer(renkuLabServer)
+      .addPath(lineage.EndpointDoc.path)
+      .addSecurity(securityScheme)
```
jachro, line 50: Would that mean that we'll have only one security scheme in the docs? It would be nice if we can add these three: oauth, private token and no auth.

knowledge-graph/src/main/scala/io/renku/knowledgegraph/docs/model.scala
```diff
@@ -0,0 +1,387 @@
...
+
+  case class Status(code: Int, name: String)
+
+  final case class SecurityRequirement(schemeName: String, scopeNames: List[String])
+  final case class SecurityScheme(name: String, `type`: TokenType, description: Option[String], in: In)
+  object SecurityScheme {
+    sealed trait SchemeType {
+      def value: String
+    }
+    final case object BearerToken extends SchemeType {
```
jachro: Should we maybe nest the various `SchemeType`s under a `SchemeType` companion?

knowledge-graph/README.md
```diff
@@ -6,19 +6,19 @@ This is a microservice which provides API for the Graph DB.
...
-| GET    | ```/knowledge-graph/entities```                                          | Returns entities filtered by the given predicates`                   |
-| GET    | ```/knowledge-graph/graphql```                                           | Returns GraphQL endpoint schema                                      |
-| POST   | ```/knowledge-graph/graphql```                                           | GraphQL query endpoint                                               |
-| GET    | ```/knowledge-graph/projects/:namespace/:name```                         | Returns details of the project with the given `namespace/name`       |
-| GET    | ```/knowledge-graph/projects/:namespace/:name/datasets```                | Returns datasets of the project with the given `path`                |
-| GET    | ```/knowledge-graph/projects/:namespace/:name/files/:location/lineage``` | Returns the lineage for a the path (location) of a file on a project |
+| GET    | ```/api/kg/datasets```                                          | Returns datasets filtered by the given predicates.                   |
```
jachro: So actually, we should leave it as it was with `/knowledge-graph`. The `/api/kg` is only for the clients of the Renku as a whole. However, on the level of the KG service, it exposes the endpoints on the `/knowledge-graph`.
The only thing I'd add is the new endpoint you added, the `GET /api/kg/spec.json` (which in fact should also be `GET /knowledge-graph/spec.json`).

PR #1026, opened 2022-06-16
chore: run jena as non-root

olevski:
We need to do this to fully run all services as non-root.

/deploy #persist


jachro:
Done as part of #1040

PR #1045, opened 2022-07-15
chore: run CA certs injection as non-root

olevski:
The latest certificates image and chart can run as non-root.

This is safer than what we had before.

Also the chart and image tags are not tracking each other. This is on purpose.

/deploy #persist


jachro (approved):
Thanks Tasko for taking care of the cert stuff on graph!

PR #1051, opened 2022-07-20
fix: api url, security

travissarles:
fix: api url, security


knowledge-graph/src/main/scala/io/renku/knowledgegraph/docs/Endpoint.scala
```diff
@@ -45,30 +46,34 @@ private class EndpointImpl[F[_]: Async](serviceVersion: ServiceVersion) extends
...
+      .addNoAuthSecurity
       .addResponsesToAll(authErrorResponses)
 
-  private lazy val localServer =
-    Server("http://localhost:{port}/{basePath}",
-           "Local server",
-           Map("port" -> Variable("8080"), "basePath" -> Variable("knowledge-graph/projects"))
+  private lazy val server: Server =
+    Server(
+      url = "knowledge-graph",
```
jachro: Do you think `knowledge-graph` would do or we'd need to have `/knowledge-graph`? I mean for the url.

jachro (approved):

PR #1055, opened 2022-07-22
chore: update to openjdk 19

olevski:
Kuba I am not sure if this is that easy (i.e. if you can swap openjdk versions without code changes).

But if you can then you will make snyk very happy. You eliminate all critical, high and medium vulnerabilities that the existing images have.

image

/deploy #persist

I guess there are two options here:
* use different base image: [eclipse-temurin:17-jre-alpine](https://hub.docker.com/layers/library/eclipse-temurin/17-jre-alpine/images/sha256-e1506ba20f0cb2af6f23e24c7f8855b417f0b085708acd9b85344a884ba77767?context=explore) looks like a very good choice as it extending `alpine:3.17` (see [here](https://github.com/adoptium/containers/blob/main/17/jre/alpine/Dockerfile.releases.full))
* either we upgrade to JDK 19(20)
* or use latest Alpine and install the necessary stuff only (see the renku-jena docker)


jachro:
Tasko, I saw this alert but I think it's a bit too early for that upgrade. There are only early access builds available for OpenJDK 19 at the moment. More to that, only earlier this week I downgraded from 18 to 17 as I was getting tons of issues when running unit tests on 18.

olevski:
Cool I will convert to draft then and you can restart this when things are more stable

jachro:
Fixed in #1228

Issue #1078, opened 2022-08-30
Named Graphs in KG


jachro:
As a Renku Graph maintainer, I'd like data in the TS to be organised in Named Graphs.

As the initial approach, creating a graph for each project and one graph for all the Person objects seem to be natural and close to the way our data is organised in other sources (in GL, each project has its own space, however, the users are common). It should also bring all the advantages described [here](https://sdsc.atlassian.net/wiki/spaces/RENKU/pages/2330591235/Structuring+the+data+in+KG+s+Triples+Store).

**Acceptance criteria:**
- [ ] provision a new TS dataset with all the data bucketed in the Graphs;
  - [x] add support to Named Graphs to the jsonld4s library;
  - [x] (PR) make TG create a new `projects` dataset in the TS on the arrival of the `TS_MIGRATION_REQUEST` event;
    - [x] the dataset to be of `TDB2` type;
    - [x] the dataset to has the same Lucene mappings as the current `renku` dataset;
    - [x] the process to verify the presence of the new dataset on every event (the same way how the other datasets are currently checked);
  - [x] (PR + release) craft a new parallel provisioning route which will push the data into the relevant `project` and `person` graphs in the new dataset in the TS;
    - [x] assess if it's better to modify the existing route so:
      - [x] the queries present in the transformation flow:
        * can work with both named and default graphs;
        * get executed on both TS datasets;
      - [x] asses if it's sensible for the route to diverge at the payloads preparation and sending steps and
        * continue to provision the old `renku` dataset as before and at the same time start provisioning the new dataset with the new named graph payloads;
        * or it's rather inconvenient, would maybe cause conditional checks and paths 
    - [x] assess if it's better to add a completely new parallel route for the `TRIPLES_GENERATED` events so:
        * it's completely independent and effectively has all the old route code initially copied;
        * there's no need to modify anything in the existing route;
        * it seems that it will be by far easier to remove the old route in the future;
        * think of how the events will be directed the new route:
          * would it be better to have a completely new event type which:
            * might be easier for the new TS transformation to generate these new events
            * prevent from re-doing all the events in the old route
            * might force creating a completely new/copy of the event producer in the EL
          * or maybe there's a check somewhere on the TG handler (or even the processor) which
            * would direct the event to the new or old or both routes (it might be easy to decide to direct an event to a route where it doesn't exist yet but might be hard to say that it should be going the route but if it's already there as it could be a new event for an existing project)
[...]

PR #1080, opened 2022-09-05
feat: added TS migration to create 'projects' dataset

jachro:
This is part of the #1078 and adds the creation of the new 'projects' dataset in the TS.

/deploy #persist

### DIFF
--- a/graph-commons/src/main/scala/io/renku/triplesstore/model.scala
+++ b/graph-commons/src/main/scala/io/renku/triplesstore/model.scala
@@ -32,7 +32,7 @@ object DatasetName extends TinyTypeFactory[DatasetName](new DatasetName(_)) with
   implicit val configReader: ConfigReader[DatasetName] = stringTinyTypeReader(DatasetName)
 }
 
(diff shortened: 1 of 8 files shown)


graph-commons/src/main/scala/io/renku/triplesstore/model.scala
```diff
@@ -32,7 +32,7 @@ object DatasetName extends TinyTypeFactory[DatasetName](new DatasetName(_)) with
   implicit val configReader: ConfigReader[DatasetName] = stringTinyTypeReader(DatasetName)
 }
 
-trait DatasetConfigFile extends StringTinyType
+trait DatasetConfigFile extends Any with StringTinyType
```
eikek: Does the `extends Any` make a practical difference (maybe for the refined lib)? Or maybe `extends AnyVal` could be an option (looking at the sub types)?

eikek (approved):
👍🏼

jachro:
> 👍🏼

Good point. I was experimenting with this trait and forgot to clean it up.

eikek (approved):

PR #1082, opened 2022-09-05
refactor: improved readability in for-comprehensions

jachro:
This PR takes advantage of the `better-monadic-for` compiler plugin and boosts the readability of for-comprehensions across the project.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -174,7 +174,8 @@ lazy val commonSettings = Seq(
   publishTo := Some(Resolver.file("Unused transient repository", file("target/unusedrepo"))),
   Compile / packageDoc / publishArtifact := false,
   Compile / packageSrc / publishArtifact := false,
-  addCompilerPlugin("org.typelevel" %% "kind-projector" % "0.13.2" cross CrossVersion.full),
+  addCompilerPlugin("com.olegpy"    %% "better-monadic-for" % "0.3.1"),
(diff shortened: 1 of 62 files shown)


build.sbt
```diff
@@ -174,7 +174,8 @@ lazy val commonSettings = Seq(
   publishTo := Some(Resolver.file("Unused transient repository", file("target/unusedrepo"))),
   Compile / packageDoc / publishArtifact := false,
   Compile / packageSrc / publishArtifact := false,
-  addCompilerPlugin("org.typelevel" %% "kind-projector" % "0.13.2" cross CrossVersion.full),
+  addCompilerPlugin("com.olegpy"    %% "better-monadic-for" % "0.3.1"),
+  addCompilerPlugin("org.typelevel" %% "kind-projector"     % "0.13.2" cross CrossVersion.full),
```
eikek, line 178: Is this `CrossVersion` setting needed? I thougt this is needed sometimes when compiling for different scala versions, but this might be wrong!

eikek (approved):
Looks good 👍🏼

build.sbt
jachro: Good catch. I took it out.

jachro: It looks like the flag is needed. I just took it out for a moment but the build failed. What is more, the docs also says it's kind of required: https://github.com/typelevel/kind-projector#using-the-plugin

PR #1083, opened 2022-09-06
refactor: TS-targeting events to write to the new projects dataset

jachro:
This PR introduces modifications to `MIN_PROJECT_INFO`, `TRIPLES_GENERATED` and `CLEAN_UP` routes (on TG). The aim of the modifications is to preserve the old functionality while introducing parallel same functionality routes for the new Named Graphed `projects` dataset.

/deploy #persist


triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsprovisioning/triplesuploading/TransformationResultsUploader.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  trait Locator[F[_]] {
+    def apply[TS <: TSVersion](implicit ev: TS): TransformationResultsUploader[F, TS]
+  }
+
+  private class LocatorImpl[F[_]](defaultGraphResultsUploader: TransformationResultsUploader[F, TSVersion.DefaultGraph],
+                                  namedGraphsResultsUploader:  TransformationResultsUploader[F, TSVersion.NamedGraphs]
+  ) extends Locator[F] {
+
+    override def apply[TS <: TSVersion](implicit ev: TS) = ev match {
```
eikek: just a thought: maybe it is possible to use two overloaded methods accepting the concrete TS - like `def apply(v: TSVersion.DefaultGraph.type) …`? Only to move the selection into compile time :). Ofc, it only works if the call sites know their concrete `TSVersion` - not sure if that applies. The selection could only happens once and then it can continue with the abstract `TransformationResultsUploader`). Maybe we could even use a "compound uploader" that works for both versions (calling both in sequence or in parallel)?

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsprovisioning/triplesuploading/TransformationResultsUploader.scala
```diff
@@ -0,0 +1,113 @@
...
+import io.renku.graph.model.entities.{Person, Project}
+import io.renku.graph.model.{GitLabApiUrl, RenkuUrl}
+import io.renku.jsonld.JsonLD.JsonLDEntity
+import io.renku.jsonld.{JsonLD, NamedGraph}
+import io.renku.jsonld.syntax._
+import io.renku.triplesgenerator.events.consumers.{ProcessingRecoverableError, TSVersion}
+import io.renku.triplesstore.{ProjectsConnectionConfig, RenkuConnectionConfig, SparqlQuery, SparqlQueryTimeRecorder}
+import org.typelevel.log4cats.Logger
+
+private trait TransformationResultsUploader[F[_], TS <: TSVersion] {
```
eikek: Does the `TS <: TSVersion` type parameter needs to be visible to users? Both methods seem not to use it and so I thought to make it an "implementation detail" by using a type member. When using this interface it can be used without knowing the concrete version the uploader is for. I think it is unified in the locator anyways. wdyt?

eikek (approved):
Nice!!

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsprovisioning/triplesuploading/TransformationResultsUploader.scala
jachro: Yes, you are right, that might be better.

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsprovisioning/triplesuploading/TransformationResultsUploader.scala
jachro: That's true, passing the TSVersion implicitly does not give us much here. I'll try to change it.

PR #1089, opened 2022-09-15
fix: Using IO.delay to defer evaluating the logging call

eikek:
When using `pure` the effect is evaluated exactly once when creating the IO, but not when evaluating it.


jachro (approved):
Thanks Eike for spotting this!

PR #1093, opened 2022-09-21
refactor: Choose how Apache Jena is provided in tests

eikek:
Improve choosing between container flavors and a locally running jena.


eikek:
This is just a proposal for an easier way of choosing where tests connect to.

PR #1095, opened 2022-09-22
chore: prepare 2.19.0

jachro:
* fix: adding --skip-template-update to renku migrate
* feat: Adding `visibility` to dataset detail response (#1092)
* refactor: Better GitLab stub (#1088)
* fix: Using IO.delay to defer evaluating the logging call (#1089)
* refactor: add generic JSON encoder for all TinyTypes (#1087)
* chore: KG processes documentation


eikek (approved):
🚀

PR #1096, opened 2022-09-23
chore: Add more detailed release instructions

eikek:
Add more detailed release instructions to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,55 @@ Hence, we are trying to find and then follow good patterns in naming, code organ
 
 #### Releasing
 
-The standard release process is done manually.
+The standard release process is done manually. There are multiple
+repositories taking part in the process. The
+[renku](https://github.com/SwissDataScienceCenter/renku) project
+contains helm charts for deploying to kubernetes and the acceptance
+tests. The
+[terraform-renku](https://github.com/SwissDataScienceCenter/terraform-renku)
+project contains deployment descriptions for all environments.
+
+**renku-graph project:**
+- create a branch off the `origin/release` branch, name it
+  `prep-<version>`
+- either cherry-pick from development to here or do a `git rebase
+  origin/development`
+- make a PR for this branch to be merged into origin/release
+  ([example](https://github.com/SwissDataScienceCenter/renku-graph/pull/1095))
+- list some points in the description for what has been changed
+- merge the PR and create a release on github
+  - release description: list the points again but organized in
+    Features, Bugfixes etc
+    ([example](https://github.com/SwissDataScienceCenter/renku-graph/releases/tag/2.19.0))
+  - The release notes are public, so don't list refactorings.
+  - this will trigger actions that:
+    - Runs acceptance tests
+    - creates a PR on the
+      [renku](https://github.com/SwissDataScienceCenter/renku) project
+      to change the version in the `requirements.yml`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,7 +65,55 @@ Hence, we are trying to find and then follow good patterns in naming, code organ
...
+  - The release notes are public, so don't list refactorings.
+  - this will trigger actions that:
+    - Runs acceptance tests
+    - creates a PR on the
+      [renku](https://github.com/SwissDataScienceCenter/renku) project
+      to change the version in the `requirements.yml`
+      ([example](https://github.com/SwissDataScienceCenter/renku/pull/2704))
+
+**renku project:**
+- create a new entry in `CHANGELOG.rst` and amend it to the PR opened
```
jachro: There's actually one more step about updating the version in the https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/renku/Chart.yaml.

eikek: Ah thanks! that is done manually and not by the bot then?
Edit: found the relevant commit, thanks!

README.md
```diff
@@ -65,7 +65,55 @@ Hence, we are trying to find and then follow good patterns in naming, code organ
...
+  by the bot
+- Merge the PR when satisfiled
```
jachro: Maybe we could also mention somewhere that in a case a PR preparing a new release is already opened (e.g. https://github.com/SwissDataScienceCenter/renku/pull/2706) the steps are slightly different:
* update the graph version in the `requirements.yaml` on that opened PR directly
* add relevant info to the CHANGELOG on the PR as well
* once the two above are done, close the autogenerated PR for graph

README.md
jachro: 👍

README.md
eikek: What would the github action do as a result of creating a release in `renku-graph`? Does it detect an existing PR if the version is the same? Otherwise we just close the obsolete PR I guess?

jachro: I cannot say for sure what's going to happen. My point was about a slightly different situation, more or less what we are currently having on the renku repo. So UI released today so an auto-generated PR was created on renku for the UI. UI simply modified the PR and turned it into `Release 0.18.0`. When we'd release, an auto PR would be generated for graph but effectively we don't need the PR as we'd rather manually update the `Release 0.18.0` as I mentioned above.

eikek: Understand, thank you. We then just close the other autogenerated PRs - I updated the notes in that regard as well.

README.md
```diff
@@ -65,7 +65,60 @@ Hence, we are trying to find and then follow good patterns in naming, code organ
...
+- either cherry-pick from development to here or do a `git rebase
+  origin/development`
+- make a PR for this branch to be merged into origin/release
+  ([example](https://github.com/SwissDataScienceCenter/renku-graph/pull/1095))
+- list some points in the description for what has been changed
+- merge the PR and create a release on github
+  - release description: list the points again but organized in
+    Features, Bugfixes etc
+    ([example](https://github.com/SwissDataScienceCenter/renku-graph/releases/tag/2.19.0))
+  - The release notes are public, so don't list refactorings.
```
jachro: A super tiny thing. This only point starts with a capital _'T'_ so let's maybe align that with the rest.

jachro (approved):
👍

PR #1100, opened 2022-09-28
feat: Start a nix development setup

eikek:
Running `nix develop` (or using direnv) creates an environment containing necessary tools for developing. The correct jdk, postgres and fuseki server is installed along with helper scripts for starting external services.


jachro (approved):
Thanks!

PR #1101, opened 2022-09-28
feat: MEMBER_SYNC flow to write to the new projects dataset

jachro:
/deploy #persist

### DIFF
--- a/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/membersync/KGSynchronizer.scala
+++ b/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/membersync/KGSynchronizer.scala
@@ -23,15 +23,18 @@ import io.renku.triplesstore.SparqlQuery
 
 private trait KGSynchronizer[F[_]] {
   def syncMembers(path: projects.Path, membersInGL: Set[GitLabProjectMember]): F[SyncSummary]
+}
 
-  protected def findMembersToAdd(membersInGitLab: Set[GitLabProjectMember],
-                                 membersInKG:     Set[KGProjectMember]
+private object KGSynchronizerFunctions {
+
+  def findMembersToAdd(membersInGitLab: Set[GitLabProjectMember],
+                       membersInKG:     Set[KGProjectMember]
   ): Set[GitLabProjectMember] = membersInGitLab.collect {
     case member @ GitLabProjectMember(gitlabId, _) if !membersInKG.exists(_.gitLabId == gitlabId) => member
   }
 
-  protected def findMembersToRemove(membersInGitLab: Set[GitLabProjectMember],
-                                    membersInKG:     Set[KGProjectMember]
+  def findMembersToRemove(membersInGitLab: Set[GitLabProjectMember],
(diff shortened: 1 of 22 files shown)


triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/membersync/KGSynchronizer.scala
```diff
@@ -0,0 +1,42 @@
...
+
+package io.renku.triplesgenerator.events.consumers.membersync
+
+import io.renku.graph.model.projects
+import io.renku.triplesstore.SparqlQuery
+
+private trait KGSynchronizer[F[_]] {
+  def syncMembers(path: projects.Path, membersInGL: Set[GitLabProjectMember]): F[SyncSummary]
+
+  protected def findMembersToAdd(membersInGitLab: Set[GitLabProjectMember],
```
eikek: Since these two methods don't seem to be related (code-wise) to this trait - what do you think about putting them in a different scope? maybe more utility like one or in some companion? (I suppose once the default graph is removed it we could move it back to the implementation?)

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/membersync/namedgraphs/UpdatesCreator.scala
```diff
@@ -0,0 +1,147 @@
...
+    val projectId = projects.ResourceId(path)
+    members
+      .map(member => generateEdge(projectId, member.resourceId))
+      .map { edge =>
+        SparqlQuery
+          .of(
+            name = "unlink project member",
+            Prefixes of schema -> "schema",
+            s"""|DELETE DATA { 
+                |  GRAPH <${GraphClass.Project.id(projectId)}> { $edge }
```
eikek, line 56: Question for curiosity/educating myself: is it possible to delete many edges "at once", in one query? (instead of generating multiple queries?)

eikek (approved):
Nice! 👍🏼

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/membersync/namedgraphs/UpdatesCreator.scala
jachro: Yes, effectively all the `DELETE`/`INSERT` queries with `DATA` suffix simply list a set of triples to delete/insert. We could actually do a single query here instead of many, however,
* Jena is generally slow with updates and sometimes it's faster to do many tiny queries than a single big one (although for the `DATA` queries there might not be not such a huge difference);
* as we don't have transactions if we have reasonably atomic updates we can try to find out which queries were done and which not.
To be true, in this particular case, there should be not much difference either way.

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/membersync/KGSynchronizer.scala
jachro: Good point. I've just extracted these two protected defs into a separate object.

PR #1105, opened 2022-10-06
refactor: Update scala to 2.13.9

eikek:
Some source changes are required:

- Infering the type `Any` is now emitting a warning. All occurences have been annotated explicitely or the type parameters have been corrected.

- Working around issue https://github.com/scala/bug/issues/12646 by moving a `var` from a `trait` into a `class`

### DIFF
--- a/acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/stubs/gitlab/GitLabStubSupport.scala
+++ b/acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/stubs/gitlab/GitLabStubSupport.scala
@@ -70,6 +70,7 @@ object GitLabStubSupport {
     def shutdown(): Unit =
       close.unsafeRunSync()
   }
+
   object GitLabStubStarter {
     def apply(gitLabUrl: GitLabUrl)(implicit rt: IORuntime, logger: Logger[IO]): GitLabStubStarter = {
       val stub = GitLabApiStub.empty[IO].unsafeRunSync()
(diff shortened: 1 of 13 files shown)


acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/stubs/gitlab/GitLabStubSupport.scala
```diff
@@ -33,26 +36,44 @@ trait GitLabStubSupport extends BeforeAndAfterAll with BeforeAndAfter with GitLa
...
+
+    def startup(): Unit = {
+      val (_, close) = stub.resource(host.value, port).allocated.unsafeRunSync()
+      this.close = close
+    }
+
+    def shutdown(): Unit =
+      close.unsafeRunSync()
+  }
+  object GitLabStubStarter {
```
jachro, line 74: Maybe a line break between the class and the object :) ?

tiny-types/src/test/scala/io/renku/tinytypes/TinyTypeSpec.scala
```diff
@@ -29,12 +29,17 @@ import org.scalatest.matchers.should
 import org.scalatest.wordspec.AnyWordSpec
 import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
 
+object TinyTypeData {
+  val arbitraryValues: List[Any] =
```
jachro, line 33: I know it's more performant to keep the values in a separate object but what do you think if we'd put the `arbitraryValues` val as a private member somewhere in the test class?

jachro (approved):
Thanks Eike. I've just left a couple of super minor comments.

tiny-types/src/test/scala/io/renku/tinytypes/TinyTypeSpec.scala
eikek: I have no strong opinion here, can do that as well. It would be duplicated in three test classes I think, which is not problematic, just saying :)

PR #1106, opened 2022-10-06
feat: TS migration to port projects from the Default to Named Graphs DS

jachro:
This PR is a part of the work described in #1078. It creates a new TS migration on the TG that finds all projects that do not exist in the Projects but are present in the Default Dataset and sends the `REDO_PROJECT_TRANSFORMATION` events for them.

### DIFF
--- a/event-log/src/test/scala/io/renku/eventlog/events/EventsFinderSpec.scala
+++ b/event-log/src/test/scala/io/renku/eventlog/events/EventsFinderSpec.scala
@@ -48,7 +48,7 @@ class EventsFinderSpec extends AnyWordSpec with IOSpec with InMemoryEventLogDbSp
 
     "return the List of events of the project the given path" in new TestCase {
       val projectId = projectIds.generateOne
-      val infos     = eventInfos(fixed(projectPath)).generateList(maxElements = 10)
+      val infos     = eventInfos(fixed(projectPath)).generateList(max = 10)
 
       infos foreach { info =>
         val eventId = CompoundEventId(info.eventId, projectId)
@@ -83,7 +83,7 @@ class EventsFinderSpec extends AnyWordSpec with IOSpec with InMemoryEventLogDbSp
 
     "return the List of events of the project with the given path and the given PagingRequest" in new TestCase {
       val projectId = projectIds.generateOne
-      val infos     = eventInfos(fixed(projectPath)).generateList(minElements = 3, maxElements = 10)
+      val infos     = eventInfos(fixed(projectPath)).generateList(min = 3, max = 10)
 
       infos foreach { info =>
         val eventId = CompoundEventId(info.eventId, projectId)
@@ -115,7 +115,7 @@ class EventsFinderSpec extends AnyWordSpec with IOSpec with InMemoryEventLogDbSp
 
     "return the List of events of the project with the given path and the given direction" in new TestCase {
       val projectId = projectIds.generateOne
-      val infos     = eventInfos(fixed(projectPath)).generateList(minElements = 3, maxElements = 10)
+      val infos     = eventInfos(fixed(projectPath)).generateList(min = 3, max = 10)
 
       infos foreach { info =>
         val eventId = CompoundEventId(info.eventId, projectId)
@@ -163,7 +163,7 @@ class EventsFinderSpec extends AnyWordSpec with IOSpec with InMemoryEventLogDbSp
 
     "return the List of events of the project with the given path and status filter" in new TestCase {
       val projectId = projectIds.generateOne
-      val infos     = eventInfos(fixed(projectPath)).generateList(minElements = 3, maxElements = 10)
+      val infos     = eventInfos(fixed(projectPath)).generateList(min = 3, max = 10)
 
       infos foreach { info =>
         val eventId = CompoundEventId(info.eventId, projectId)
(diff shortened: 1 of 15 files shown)


event-log/src/test/scala/io/renku/eventlog/events/EventsFinderSpec.scala
```diff
@@ -83,7 +83,7 @@ class EventsFinderSpec extends AnyWordSpec with IOSpec with InMemoryEventLogDbSp
 
     "return the List of events of the project with the given path and the given PagingRequest" in new TestCase {
       val projectId = projectIds.generateOne
-      val infos     = eventInfos(fixed(projectPath)).generateList(minElements = 3, maxElements = 10)
+      val infos     = eventInfos(fixed(projectPath)).generateList(min = 3, max = 10)
```
eikek, line 86: ❤️

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToNamedGraphs.scala
```diff
@@ -0,0 +1,175 @@
...
+
+  override def run(): EitherT[F, ProcessingRecoverableError, Unit] = EitherT {
+    Stream
+      .iterate(1)(_ + 1)
+      .evalMap(findDefaultGraphProjects)
+      .takeThrough(_.nonEmpty)
+      .flatMap(in => Stream.emits(in))
+      .evalFilterNot { case (projectId, _) => checkExists(projectId) }
+      .map(toRedoEvent)
+      .evalMap { case (payload, ctx) => sendEvent(payload, ctx) }
```
eikek, line 64: I think one could also write `.evalMap(sendEvent.tupled)` to spare the redundant param passing - but just nitpicking here, not important :)

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToNamedGraphs.scala
```diff
@@ -0,0 +1,175 @@
...
+}
+
+private class DefaultGraphProjectsFinderImpl[F[_]](recordsFinder: RecordsFinder[F])
+    extends DefaultGraphProjectsFinder[F]
+    with Schemas {
+
+  import ResultsDecoder._
+  import io.renku.tinytypes.json.TinyTypeDecoders._
+
+  private val chunkSize: Int = 30
```
eikek: Do you think we could increase this? It's "only" id and path what is loaded, right?

triples-generator/src/test/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToNamedGraphsSpec.scala
```diff
@@ -0,0 +1,203 @@
...
+
+    def givenProject(project: ProjectInfo, existsInNamedGraphsResponse: IO[Boolean]): Unit = {
+      (namedGraphsProjectFinder.checkExists _)
+        .expects(project._1)
+        .returning(existsInNamedGraphsResponse)
+      ()
+    }
+
+    def toRedoEvent(project: ProjectInfo) = project -> EventRequestContent.NoPayload {
+      json"""{
```
eikek, line 121: Wouldn't it be beneficial if we would have a case class representing the payload of these events instead of writing the literal json each time?

eikek (approved):
Looks great! 👍🏼

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToNamedGraphs.scala
jachro: It looks like the problem is that the `sendEvent` def is overloaded. When I even try to turn the def into a function with `_` the compiler gets confused.

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToNamedGraphs.scala
jachro: Yes, we could. So far the value is absolutely arbitrary and simply on a very, very defensive side.

jachro: Raised it to 50 :)

triples-generator/src/test/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToNamedGraphsSpec.scala
jachro: Right, that's a bit more complicated thing as it goes beyond a single service. I mean, we can think if we'd like to introduce some special module that lists event payload classes or do it in some other way as I think it's rather not the best idea to simply make EL a compile-time dependency of TG.
Let's maybe leave it for now and discuss maybe at some other time.

eikek: sure, leave it for now - I was wondering more in general, unrelated to this pr. 👍🏼

PR #1107, opened 2022-10-06
refactor: TSClientImpl renamed to TSClient

jachro:
This PR simply renames a not so well named `TSClientImpl` class to `TSClient`.


eikek (approved):
👍🏼

PR #1138, opened 2022-10-12
feat: find tags using named graph datasets

eikek:
Changes the query to find tags using the named graph datasets

### DIFF
--- a/knowledge-graph/src/main/scala/io/renku/knowledgegraph/projects/datasets/tags/TagsFinder.scala
+++ b/knowledge-graph/src/main/scala/io/renku/knowledgegraph/projects/datasets/tags/TagsFinder.scala
@@ -22,8 +22,11 @@ import Endpoint.Criteria
 import cats.NonEmptyParallel
 import cats.effect.Async
 import cats.syntax.all._
+import io.renku.graph.model.RenkuUrl
+import io.renku.graph.model.projects.ResourceId
+import io.renku.graph.model.views.RdfResource
 import io.renku.http.rest.paging.{Paging, PagingResponse}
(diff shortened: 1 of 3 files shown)


knowledge-graph/src/main/scala/io/renku/knowledgegraph/projects/datasets/tags/TagsFinder.scala
```diff
@@ -59,15 +59,16 @@ private class TagsFinderImpl[F[_]: Async: NonEmptyParallel: Logger: SparqlQueryT
...
-        |  ?projId a schema:Project;
-        |          renku:projectPath '${criteria.projectPath}';
-        |          renku:hasDataset ?dsId.
-        |  ?dsId renku:slug '${criteria.datasetName}';
-        |        schema:identifier ?dsIdentifier.
-        |  ?eventId schema:about/schema:url ?dsId;
-        |           schema:name ?name;
-        |           schema:startDate ?startDate.
-        |  OPTIONAL { ?eventId schema:description ?maybeDesc }
+        |  Graph ?g { ?projId renku:projectPath '${criteria.projectPath}' }
```
jachro: Yes, things will work that way but I've noticed that reaching directly the proper graph might be a bit better. In order to do so, you can construct the Project ResourceId and use it to select the graph. An example might be this: https://github.com/SwissDataScienceCenter/renku-graph/blob/30bba7f33b692ffe41a96f153c56d58f705208e6/knowledge-graph/src/main/scala/io/renku/knowledgegraph/projects/details/KGProjectFinder.scala#L59

eikek: Thank you! I didn't realize I can create a resource-id from a path!

jachro (approved):

PR #1139, opened 2022-10-12
feat: dataset APIs to work against the Named Graphs dataset

jachro:
This PR switches the Dataset Search and Dataset Details APIs to fetch data from the Named Graphs dataset in the TS.

### DIFF
--- a/knowledge-graph/src/main/scala/io/renku/knowledgegraph/datasets/DatasetSearchResult.scala
+++ b/knowledge-graph/src/main/scala/io/renku/knowledgegraph/datasets/DatasetSearchResult.scala
@@ -19,7 +19,7 @@
 package io.renku.knowledgegraph
 package datasets
 
-import DatasetSearchResult.ProjectsCount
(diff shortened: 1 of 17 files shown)


knowledge-graph/src/main/scala/io/renku/knowledgegraph/datasets/DatasetSearchResult.scala
```diff
@@ -32,48 +32,48 @@ import io.renku.tinytypes.constraints.NonNegativeInt
...
-                             maybeDescription,
-                             creators,
-                             date,
-                             exemplarProjectPath,
-                             projectsCount,
-                             keywords,
-                             images
-        ) =>
-      json"""{
+  final case class ExemplarProject(id: projects.ResourceId, path: projects.Path)
```
eikek, line 49: For curiosity: why is it called "Exemplar" - it sounds a bit like being test data :) Or is this the example for the api docs?

eikek (approved):
Great 👍🏼

PR #1177, opened 2022-11-08
refactor: removed code provisioning to Default Graph in the TS

### LINKED ISSUE #1078: Named Graphs in KG

jachro:
This is the last clean-up PR from the migration to the Named Graphs feature.

closes #1078


eikek (approved):
👍🏼

PR #1194, opened 2022-11-16
feat: KG to work with Project Access Tokens

jachro:
This PR introduces the concept of Project Access Tokens.

### DIFF
--- a/graph-commons/src/main/scala/io/renku/http/client/AccessToken.scala
+++ b/graph-commons/src/main/scala/io/renku/http/client/AccessToken.scala
@@ -37,10 +37,10 @@ object AccessToken {
       extends TinyTypeFactory[PersonalAccessToken](new PersonalAccessToken(_))
       with NonBlank[PersonalAccessToken]
 
-  final class OAuthAccessToken private (val value: String) extends AnyVal with UserAccessToken
-  object OAuthAccessToken
-      extends TinyTypeFactory[OAuthAccessToken](new OAuthAccessToken(_))
-      with NonBlank[OAuthAccessToken]
+  final class UserOAuthAccessToken private (val value: String) extends AnyVal with UserAccessToken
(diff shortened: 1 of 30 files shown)


graph-commons/src/main/scala/io/renku/http/client/AccessToken.scala
```diff
@@ -28,40 +28,60 @@ import java.util.Base64
...
+      extends TinyTypeFactory[ProjectAccessToken](new ProjectAccessToken(_))
+      with NonBlank[ProjectAccessToken]
+
   private val base64Decoder = Base64.getDecoder
   private val base64Encoder = Base64.getEncoder
 
   implicit val accessTokenEncoder: Encoder[AccessToken] = {
+    case ProjectAccessToken(token) =>
+      Json.obj("projectAccessToken" -> Json.fromString(new String(base64Encoder.encode(token.getBytes(UTF_8)), UTF_8)))
     case OAuthAccessToken(token) =>
```
eikek: Now the name `OAuth…` seems a bit too generic. It is only for a user token used. If understood correctly the "project token" is an oauth token as well (being used in an oauth flow)? (nit picking obviously :))

eikek (approved):
👍🏼

graph-commons/src/main/scala/io/renku/http/client/AccessToken.scala
jachro: True. Perhaps, it'd be better to name it `UserOAuthAccessToken`? What do you think?

eikek: Yes, that sounds good 👍🏼

jachro: 👍🏻

PR #1197, opened 2022-11-17
feat: DB migration adding expiry and created dates

jachro:
This PR adds two new nullable, timestampz columns to the `projects_tokens` table: `expiry_date` and `created_at`

### DIFF
--- a/token-repository/src/test/scala/io/renku/tokenrepository/repository/InMemoryProjectsTokensDb.scala
+++ b/token-repository/src/test/scala/io/renku/tokenrepository/repository/InMemoryProjectsTokensDb.scala
@@ -21,8 +21,9 @@ package io.renku.tokenrepository.repository
 import cats.data.Kleisli
 import cats.effect.IO
 import com.dimafeng.testcontainers._
-import io.renku.db.{PostgresContainer, SessionResource}
+import io.renku.db.PostgresContainer
 import io.renku.testtools.IOSpec
+import io.renku.tokenrepository.repository.ProjectsTokensDB.SessionResource
 import natchez.Trace.Implicits.noop
 import org.scalatest.Suite
 import skunk._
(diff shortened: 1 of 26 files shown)


token-repository/src/test/scala/io/renku/tokenrepository/repository/InMemoryProjectsTokensDb.scala
```diff
@@ -47,16 +48,15 @@ trait InMemoryProjectsTokensDb extends ForAllTestContainer {
...
-    sessionResource
-      .useK {
-        val query: Query[Void, Boolean] = sql"""select exists (select * from projects_tokens);""".query(bool)
-        Kleisli[IO, Session[IO], Option[Boolean]](_.option(query).recover { case _ => None })
-      }
-      .unsafeRunSync()
-      .isDefined
+  protected def tableExists(): Boolean = SessionResource[IO]
+    .useK {
+      val query: Query[Void, Boolean] = sql"""select exists (select * from projects_tokens);""".query(bool)
```
eikek: we could also query the catalog. or use a simpler query like `select 1 from project_tokens`? (sorry, I know it's just re-arranging code)

eikek: … or if i think about it again, this might both return `false` if the table exists but is empty, no? maybe `select max(1) from project_tokens` should work

eikek (approved):
I'm probably a bit too blind right now, but I can't find any reference to the two mentioned columns in the code? 🤔

token-repository/src/test/scala/io/renku/tokenrepository/repository/InMemoryProjectsTokensDb.scala
jachro: Yes, I changed the query to look into the `pg_tables`.

PR #1199, opened 2022-11-18
feat: include plan-type info into cross entity search

eikek:
When querying workflows an additional attribute indicates whether it is a "normal" workflow or a composite.

### DIFF
--- a/knowledge-graph/src/main/scala/io/renku/knowledgegraph/entities/model.scala
+++ b/knowledge-graph/src/main/scala/io/renku/knowledgegraph/entities/model.scala
@@ -180,10 +180,10 @@ private object model {
       }
       object WorkflowType {
         case object Composite extends WorkflowType
-        case object Basic     extends WorkflowType
+        case object Step      extends WorkflowType
 
         val all: NonEmptyList[WorkflowType] =
-          NonEmptyList.of(Composite, Basic)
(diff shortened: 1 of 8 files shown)


knowledge-graph/src/main/scala/io/renku/knowledgegraph/entities/model.scala
```diff
@@ -166,13 +167,31 @@ private object model {
...
+        case object Basic     extends WorkflowType
+
+        val all: NonEmptyList[WorkflowType] =
+          NonEmptyList.of(Composite, Basic)
+
+        def fromName(str: String): Either[String, WorkflowType] =
+          all.find(_.name.equalsIgnoreCase(str)).toRight(s"Invalid workflowType name: $str")
+
+        implicit val encoder: Encoder[WorkflowType] =
+          Encoder.encodeString.contramap(_.name)
```
eikek, line 192: Here I just choose a simple enum (string) for transporting the value to the consumer of the api. But open for other ways :)-

jachro: Sure, I'd maybe name it `Step` instead of `Basic` for better consistency in naming.

jachro (approved):

PR #1204, opened 2022-11-23
feat: event payload endpoint

eikek:
- Adds an endpoint for returning the event payload given an event id and project path
- Refactors db migrations used in tests to simply use the production db migrations and removed an obsolete migration that was left in the test sources


jachro (approved):
Great work! Thanks Eike!

PR #1208, opened 2022-11-25
feat: DB migration replacing existing non-PAT to PAT tokens

jachro:
This PR adds a DB migration replacing existing non-Project Access Tokens with Project Access Tokens. It also makes the DB Migrations run asynchronously.

### DIFF
--- a/token-repository/src/main/scala/io/renku/tokenrepository/repository/association/ProjectAccessTokenCreator.scala
+++ b/token-repository/src/main/scala/io/renku/tokenrepository/repository/association/ProjectAccessTokenCreator.scala
@@ -24,11 +24,12 @@ import com.typesafe.config.{Config, ConfigFactory}
 import io.renku.graph.model.projects
 import io.renku.http.client.AccessToken.ProjectAccessToken
 import io.renku.http.client.{AccessToken, GitLabClient}
+import org.http4s.Status.{Forbidden, NotFound}
 
 import java.time.{LocalDate, Period}
 
 private[tokenrepository] trait ProjectAccessTokenCreator[F[_]] {
-  def createPersonalAccessToken(projectId: projects.Id, accessToken: AccessToken): F[TokenCreationInfo]
+  def createPersonalAccessToken(projectId: projects.Id, accessToken: AccessToken): F[Option[TokenCreationInfo]]
 }
 
 private[tokenrepository] object ProjectAccessTokenCreator {
@@ -59,7 +60,9 @@ private class ProjectAccessTokenCreatorImpl[F[_]: Async: GitLabClient](
   import org.http4s.implicits._
   import org.http4s.{EntityDecoder, Request, Response, Status}
 
-  override def createPersonalAccessToken(projectId: projects.Id, accessToken: AccessToken): F[TokenCreationInfo] = {
+  override def createPersonalAccessToken(projectId:   projects.Id,
+                                         accessToken: AccessToken
+  ): F[Option[TokenCreationInfo]] = {
     val payload = json"""{
       "name":       "renku",
       "scopes":     ["api", "read_repository"],
@@ -70,8 +73,9 @@ private class ProjectAccessTokenCreatorImpl[F[_]: Async: GitLabClient](
     )(accessToken.some)
   }
 
-  private lazy val mapResponse: PartialFunction[(Status, Request[F], Response[F]), F[TokenCreationInfo]] = {
-    case (Created, _, response) => response.as[TokenCreationInfo]
+  private lazy val mapResponse: PartialFunction[(Status, Request[F], Response[F]), F[Option[TokenCreationInfo]]] = {
+    case (Created, _, response)       => response.as[TokenCreationInfo].map(Option.apply)
+    case (Forbidden | NotFound, _, _) => Option.empty[TokenCreationInfo].pure[F]
   }
 
   private implicit lazy val tokenDecoder: EntityDecoder[F, TokenCreationInfo] = {
--- a/token-repository/src/main/scala/io/renku/tokenrepository/repository/association/ProjectAccessTokenCreator.scala
+++ b/token-repository/src/main/scala/io/renku/tokenrepository/repository/association/ProjectAccessTokenCreator.scala
@@ -27,11 +27,11 @@ import io.renku.http.client.{AccessToken, GitLabClient}
 
 import java.time.{LocalDate, Period}
 
-private trait ProjectAccessTokenCreator[F[_]] {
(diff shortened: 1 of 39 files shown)


token-repository/src/main/scala/io/renku/tokenrepository/repository/association/ProjectAccessTokenCreator.scala
```diff
@@ -70,8 +73,9 @@ private class ProjectAccessTokenCreatorImpl[F[_]: Async: GitLabClient](
     )(accessToken.some)
   }
 
-  private lazy val mapResponse: PartialFunction[(Status, Request[F], Response[F]), F[TokenCreationInfo]] = {
-    case (Created, _, response) => response.as[TokenCreationInfo]
+  private lazy val mapResponse: PartialFunction[(Status, Request[F], Response[F]), F[Option[TokenCreationInfo]]] = {
+    case (Created, _, response)       => response.as[TokenCreationInfo].map(Option.apply)
+    case (Forbidden | NotFound, _, _) => Option.empty[TokenCreationInfo].pure[F]
```
eikek, line 78: just wondering if we have some logging in case of error (maybe via some middleware)?

token-repository/src/main/scala/io/renku/tokenrepository/repository/init/ExpiryAndCreatedDatesNotNull.scala
```diff
@@ -0,0 +1,47 @@
...
+import cats.effect.Spawn
+import cats.syntax.all._
+import io.renku.tokenrepository.repository.ProjectsTokensDB.SessionResource
+import org.typelevel.log4cats.Logger
+
+private object ExpiryAndCreatedDatesNotNull {
+  def apply[F[_]: Spawn: Logger: SessionResource]: DBMigration[F] = new ExpiryAndCreatedDatesNotNull[F]
+}
+
+private class ExpiryAndCreatedDatesNotNull[F[_]: Spawn: Logger: SessionResource] extends DBMigration[F] {
```
eikek: Do we need the `Spawn` here?

token-repository/src/main/scala/io/renku/tokenrepository/repository/init/TokensMigrator.scala
```diff
@@ -0,0 +1,159 @@
...
+  import skunk.implicits._
+  import tokenCrypto._
+  import tokenRemover._
+  import tokenValidator._
+
+  override def run(): F[Unit] =
+    Stream
+      .emit(())
+      .repeat
+      .evalMap(_ => findTokenWithoutDates)
```
eikek: I think these 3 can be written as `Stream.repeatEval(findTokenWithoutDates)` - but just thinking if it wouldn't be nicer to have skunk return a stream of results already? that would involve only one sql query and a stream would be built around the cursor. or does it interfere with the other statements that are run on each project?

token-repository/src/main/scala/io/renku/tokenrepository/repository/init/TokensMigrator.scala
```diff
@@ -0,0 +1,159 @@
...
+          }
+      )
+      .arguments(Void)
+
+  private def decryptOrDelete(project: Project, encToken: EncryptedAccessToken): F[Option[(Project, AccessToken)]] =
+    decrypt(encToken)
+      .map(t => Option(project -> t))
+      .recoverWith { case ex: Throwable =>
+        Logger[F].error(ex)(show"$logPrefix $project token decryption failed; deleting") >>
+          delete(project.id) >>
```
eikek: Is the project deleted here or the token?

token-repository/src/main/scala/io/renku/tokenrepository/repository/init/TokensMigrator.scala
```diff
@@ -0,0 +1,159 @@
...
+  import tokenRemover._
+  import tokenValidator._
+
+  override def run(): F[Unit] =
+    Stream
+      .emit(())
+      .repeat
+      .evalMap(_ => findTokenWithoutDates)
+      .takeThrough(_.nonEmpty)
+      .flatMap(maybeProjectAndToken => Stream.emits(maybeProjectAndToken.toList))
```
eikek: I think this could be same as `.unNoneTerminate` (stopping on first `None`)? (there also is `.unNone` to just filter out all `None`s)

eikek (approved):
👍🏼 👍🏼 👍🏼

token-repository/src/main/scala/io/renku/tokenrepository/repository/association/ProjectAccessTokenCreator.scala
jachro: Yes, effectively all the cases not caught in the mapping are turned into exceptions and logged where we deal with the effects.

token-repository/src/main/scala/io/renku/tokenrepository/repository/init/ExpiryAndCreatedDatesNotNull.scala
jachro: Good catch. `MonadCancelThrow` is enough.

token-repository/src/main/scala/io/renku/tokenrepository/repository/init/TokensMigrator.scala
jachro: > `Stream.repeatEval(findTokenWithoutDates)`
Perfect! Thanks!

Yes, I was thinking about the cursor here. The only reason why I didn't go for that is that the process involves several calls to GitLab which can potentially take some time or can break. With the single query approach, each found row is somehow dealt with autonomously and we don't need to keep the cursor open for long. I know it's much slower and less efficient but I thought the speed was not a problem here.

token-repository/src/main/scala/io/renku/tokenrepository/repository/init/TokensMigrator.scala
jachro: The token. I'll add some context here for better readability.

token-repository/src/main/scala/io/renku/tokenrepository/repository/init/TokensMigrator.scala
jachro: Perfect! It's so nice to have fs2 skilled eye on the team ;)

token-repository/src/main/scala/io/renku/tokenrepository/repository/init/TokensMigrator.scala
eikek: Ah right! makes total sense to not have a connection open when doing the work

PR #1214, opened 2022-11-30
feat: Create client for event-log microservice

eikek:
Moves model classes from event-log to graph-commons where already some of them were defined. Remove `CompoundId` because it is not used.


graph-commons/src/main/scala/io/renku/graph/model/events.scala
```diff
@@ -220,6 +247,65 @@ object events {
...
+          id              <- cursor.downField("id").as[EventId]
+          path            <- cursor.downField("projectPath").as[projects.Path]
+          status          <- cursor.downField("status").as[EventStatus]
+          processingTimes <- cursor.downField("processingTimes").as[List[StatusProcessingTime]]
+          date            <- cursor.downField("date").as[EventDate]
+          executionDate   <- cursor.downField("executionDate").as[ExecutionDate]
+        } yield EventInfo(id, path, status, date, executionDate, None, processingTimes)
+      }
+
+    implicit val jsonEncoder: Encoder[EventInfo] = eventInfo => {
```
jachro: One question/concern. I think we talked about that already but I'm still not convinced about the JSON encoder belonging here. If this is supposed to be a shared code for event-log clients I think the better place for the encoder might somewhere nearby the endpoint. Any thoughts?

eikek: I'm also fine to move it. For me it is ok to put them next to each other. The reason is that this case class defines what data can be consumed from the endpoint and here there is a "canonical" json format for it. So I would put the decoder and encoder close to each other, because they are supposed to complement each other. But not a strong opinion. Where would you prefer the encoder to be defined? Maybe as a private value in the `EventsEndpoint` companion?

jachro: Yes, for me the endpoint seems to be probably a better place.

jachro (approved):

PR #1215, opened 2022-11-30
feat: token refresh process

### LINKED ISSUE #1193: KG to use Project Access Tokens
Acceptance criteria:
- [x] (PR) add a new Project Access Token (PAT) variant of AccessToken;
  - [x] adapt places where an AccessToken is used
- [x] (PR) DB migration adding nullable `expiry_date` and `created_at`
- [x] (PR) change the behaviour of `PUT /projects/:id/tokens`
  - [x] check if PAT is already stored in the DB;
    - [x] if yes and it's valid (do a validation call (`HEAD https://gitlab.dev.renku.ch/api/v4/user` returns `200`)) do nothing
    - [x] else
      - [x] generate a PAT using the received oauth/personal token
        - [x] `POST https://gitlab/api/v4/projects/:projectId/access_tokens` with payload
        ```json
        { "name":"renku", "scopes":["api", "read_repository"] , "expires_at" : "now + 1y"}
        ```
      - [x] persist the PAT setting the two new dates
- [x] (PR) DB migration replacing existing non-PAT
  - [x] make the migrations to run async;
  - [x] for each token stored in the DB without `expiry_date`
    - [x] generate a PAT using the found oauth/personal token
      - [x] if generation succeeded persist the PAT
      - [x] else 
        - [x] for connectivity issues retry forever
        - [x] remove the invalid token from DB
      - [x] make `expiry_date` and `created_at` `NOT NULL`
- [ ] (PR) create a process to recreate tokens that are about to expire
  - [ ] create a process on the token-repository service that finds out all PATs that expiries in one day from the date it's checked;
  - [ ] for each found token generate a new PAT using the old one
  - [ ] persist the new PAT with the two new dates
  - [ ] keep retrying if generation failed due to connectivity issues
  - [ ] remove the existing token from DB in other cases (the token is invalid)
  - [ ] remove all the tokens from GL that have `renku` name and their expiration date is before or equal today + one day

jachro:
This PR introduces the Project Access Token refresh process

closes #1193

### DIFF
--- a/graph-commons/src/main/scala/io/renku/http/tinytypes/TinyTypeURIEncoder.scala
+++ b/graph-commons/src/main/scala/io/renku/http/tinytypes/TinyTypeURIEncoder.scala
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2022 Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.renku.http.tinytypes
+
+import io.renku.tinytypes.TinyType
+import org.http4s.Uri
+import org.http4s.Uri.Path.SegmentEncoder
+
+object TinyTypeURIEncoder extends TinyTypeURIEncoder
+
+// It looks it might be convenient to have this functionality available out of the box for all TinyTypes.
+// However, the tiny-types module does not depend on http4s so maybe additional module (like tiny-types-http4s)
+// could be created at some point.
+trait TinyTypeURIEncoder {
+
+  implicit def segmentEncoder[TT <: TinyType](implicit
+      valueEncoder: SegmentEncoder[TT#V]
+  ): Uri.Path.SegmentEncoder[TT] = valueEncoder.contramap(_.value)
+}
--- a/token-repository/src/main/scala/io/renku/tokenrepository/repository/refresh/EventsFinder.scala
+++ b/token-repository/src/main/scala/io/renku/tokenrepository/repository/refresh/EventsFinder.scala
@@ -65,6 +65,6 @@ private class EventsFinderImpl[F[_]: MonadCancelThrow: SessionResource](
             TokenCloseExpiration(Project(id, path), token)
           }
       )
-      .arguments(LocalDate.now() plusDays 1)
+      .arguments(CloseExpirationDate().value)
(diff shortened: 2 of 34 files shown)


graph-commons/src/main/scala/io/renku/http/tinytypes/TinyTypeURIEncoder.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import io.renku.tinytypes.TinyType
+import org.http4s.Uri
+import org.http4s.Uri.Path.SegmentEncoder
+
+object TinyTypeURIEncoder extends TinyTypeURIEncoder
+
+// It looks it might be convenient to have this functionality available out of the box for all TinyTypes.
+// However, the tiny-types module does not depend on http4s so maybe additional module (like tiny-types-http4s)
+// could be created at some point.
```
eikek, line 29: I agree 👍🏼 , would be beneficial for us as we use http4s and it could be the home for other http4s related utilites (query param encoder/decoder etc)

token-repository/src/main/scala/io/renku/tokenrepository/repository/refresh/EventsFinder.scala
```diff
@@ -0,0 +1,69 @@
...
+import ProjectsTokensDB.SessionResource
+import association.Project
+import cats.effect.MonadCancelThrow
+import io.renku.db.{DbClient, SqlStatement}
+import io.renku.graph.model.projects
+import io.renku.metrics.LabeledHistogram
+
+import java.time.LocalDate
+
+private trait EventsFinder[F[_]] {
```
eikek: Probably miss something: I don't understand why it is a "events finder" when it queries project tokens?

token-repository/src/main/scala/io/renku/tokenrepository/repository/refresh/ExpiredTokensFinder.scala
```diff
@@ -0,0 +1,78 @@
...
+    jsonOf[F, List[TokenInfo]]
+  }
+
+  private lazy val renkuTokens: TokenInfo => Boolean = { case (_, name, _) =>
+    name == renkuTokenName
+  }
+
+  private lazy val expired: TokenInfo => Boolean = {
+    case (_, _, None)         => true
+    case (_, _, Some(expiry)) => (expiry.value compareTo CloseExpirationDate().value) <= 0
```
eikek, line 76: minor nitpick: there is a side effect of getting the current day. It also uses the system time zone, which is probably fine for us. It could have some influence when comparing for one-day differences depending on what timezone the other local date is.

eikek (approved):
👍🏼 Nice!!

token-repository/src/main/scala/io/renku/tokenrepository/repository/refresh/EventsFinder.scala
jachro: True, I renamed it to `TokensToRefreshFinder`

token-repository/src/main/scala/io/renku/tokenrepository/repository/refresh/ExpiredTokensFinder.scala
jachro: I see your point but I think we should be fine. The reason is that we pass the expiry date to GL on token creation (using token-repository's time zone), so even if Postgres or GL are running in different time zones we don't depend on them. At least that's how I see the problem and solution. Generally, I think there should be no place in our services where we'd rely on date/time from Postgres container. I mean, there should be no SQL query where `now()` is used as we always pass the date/time from scala code.

PR #1218, opened 2022-12-01
feat: Add ts migration for composite plans

### LINKED ISSUE #721: Add support for the CompositePlan entity
As a renku graph user, I'd like all flavours of `Plan` entities to be present in the KG.

**Acceptance criteria:**
- [x] existing `Plan` entity to get a new `derivedFrom` property
  - [x] the property to be encoded/decoded without the `schema:URL` object as it's currently done on `Dataset` entity;
  - [x] there has to be a `Plan` entity which `derivedFrom` points to within metadata of the `Project`; broken links should make `Plan` instantiation fail;
  - [x] `Plan` should get additional `renku:Plan` type;
  - [x] `Plan` to be decoded based on the `renku:Plan` type;
  - [x] add validation of `invalidationTime`;
- [x] add support for `CompositePlan` entity (for more details see [here](https://github.com/SwissDataScienceCenter/renku-python/blob/cd967734168f36ea1fab7d3ff4e8d31261adfc24/renku/command/schema/composite_plan.py))
  - [x] `CompositePlan` to be distinguishable from `Plan` by the `renku: CompositePlan`
  - [x] probably a common parent `Plan` trait would be worth introducing so both types of `Plan` can be used in `Project` or `Association`
  - [x] all `mappings` has to be pointing to an existing parameter on the `Plan` where it's defined or on its children;
    - [x] validation if all mapsTo are not broken and point to parameters of relevant types; 
  - [x] child `plans` can be either `Plan` or `CompositePlan`;
  - [x] `links` has to be referencing existing parameters on the `Plan` where it's defined or on its children;
    - [x] validation if all `source` and `sinks` are not broken and point to parameters of a relevant type;
  - [x] add a test proving deserialization of a Plan and a CompositePlan existing in one `hasPlan` property works;
  - [x] the `CompositePlan` is immutable so no need to special transformation operations;
- [x] CompositePlans should be visible for the Cross-Entity search;
- [ ] write a TS migration that will cause re-provisioning all the projects having at least one `CompositePlan`
  - [x] EL to expose a new endpoint `GET /events/:event-id/:project-path/payload` which would return payload associated with the event
    - [x] 200 if payload exists for the event
    - [x] 404 if either event or payload does not exist
  - [ ] create a new TS migration that:
    - [x] for each `Project` in the TS
    - [x] call `GET /events?project-path=:path&per_page=1`
      - [ ] if there are no events or the latest event is in status `New`, `Skipped`, `GeneratingTriples`, `GenerationRecoverableFailure`, `TriplesGenerated`, `TransformingTriples`, `TransformationRecoverableFailure`, `AwaitingDeletion` or `Deleting` then do nothing
      - [ ] if the latest event is in any other status
        - [ ] extract the event_id and project_id and call `GET /events/:event-id/:project-path/payload` to fetch the payload
        - [ ] if we got `404` or we got `200` and the payload contains `CompositePlan` entity (text search could do)
        - [ ] send `POST /events` (`EventSender`)

      ```json
      {
        "categoryName": "EVENTS_STATUS_CHANGE",
        "project": {
          "id":   12,
        },
        "newStatus": "NEW"
      }
      ```

NOTE: check if the payload will be accepted by EL without `project.id`; if not, the adapt the EL so it's possible

eikek:
Adds a migration to change the status to NEW for projects to read in their composite plan. Projects are skipped that are known to not include a composite plan.

/deploy #persist

closes #721

### DIFF
--- a/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/CompositePlanProvision.scala
+++ b/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/CompositePlanProvision.scala
@@ -1,0 +1,191 @@
+/*
+ * Copyright 2022 Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.renku.triplesgenerator.events.consumers.tsmigrationrequest.migrations
+
+import cats.Show
+import cats.data.{EitherT, OptionT}
+import cats.effect._
+import cats.kernel.Monoid
+import cats.syntax.all._
+import fs2.Stream
+import io.circe.Json
+import io.circe.syntax._
+import io.renku.compression.Zip
+import io.renku.events.{CategoryName, EventRequestContent}
+import io.renku.events.producers.EventSender
+import io.renku.graph.config.EventLogUrl
+import io.renku.graph.eventlog.EventLogClient
+import io.renku.graph.eventlog.EventLogClient.{EventPayload, SearchCriteria}
+import io.renku.graph.model.events.{EventInfo, EventStatus}
+import io.renku.graph.model.projects.{Path => ProjectPath}
+import io.renku.metrics.MetricsRegistry
+import io.renku.triplesgenerator.events.consumers.ProcessingRecoverableError
+import io.renku.triplesgenerator.events.consumers.tsmigrationrequest.Migration
+import io.renku.triplesgenerator.events.consumers.tsmigrationrequest.migrations.CompositePlanProvision.{Context, Result}
+import io.renku.triplesgenerator.events.consumers.tsmigrationrequest.migrations.tooling.{AllProjects, MigrationExecutionRegister, RecoverableErrorsRecovery, RegisteredMigration}
+import io.renku.triplesstore.SparqlQueryTimeRecorder
+import org.typelevel.log4cats.Logger
+
+private[migrations] class CompositePlanProvision[F[_]: Sync: Logger](
+    override val name: Migration.Name,
+    allProjects:       AllProjects[F],
+    eventSender:       EventSender[F],
+    eventLogClient:    EventLogClient[F],
+    executionRegister: MigrationExecutionRegister[F],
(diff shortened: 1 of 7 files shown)


triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/CompositePlanProvision.scala
```diff
@@ -0,0 +1,184 @@
...
+        .evalMap(CompositePlanProvision.createContext(getEventPayload))
+        .evalMap {
+          case Some(ctx) => sendStatusChange(ctx).as(Result.sendEvent)
+          case None      => Result.skip.pure[F]
+        }
+        .compile
+        .foldMonoid
+        .flatMap(result => Logger[F].info(result.show))
+        .attempt
+        .map(_.leftMap(LogWorthyRecoverableError(show"Error while running ts migration: $name", _)))
```
eikek: Regarding errors, I wasn't quite sure. The only expected errors are when EL is down or its db migrations are running. This could theoretically happen in between. So when this would be restarted it should skip already done projects right? (because the latest event would be in state NEW). I think in general, it could be retried in this case?

jachro: I think you could simply rely on `recoverWith(recoveryStrategy.maybeRecoverableError[F, Unit])`. It contains a common way of dealing with Exceptions.

triples-generator/src/test/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/tooling/AllProjectsSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+  implicit val gitlabUrl: GitLabApiUrl = EntitiesGenerators.gitLabApiUrl
+  implicit val logger:    Logger[IO]   = TestLogger[IO]()
+  implicit lazy val timeRecorder: SparqlQueryTimeRecorder[IO] =
+    SparqlQueryTimeRecorder(TestMetricsRegistry[IO]).unsafeRunSync()
+
+  def allProjects: AllProjects[IO] = AllProjects[IO](projectsDSConnectionInfo)
+
+  "find" should {
+    "find all projects" in {
+      val projectGen = EntitiesGenerators.renkuProjectEntities(EntitiesGenerators.visibilityPublic)
```
jachro: I guess we could even go for `EntitiesGenerators.anyProjectEntities` as we should do all projects.

triples-generator/src/test/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/tooling/AllProjectsSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+
+      val projects = Gen.listOfN(32, projectGen).generateOne
```
jachro: As an option, there's also `generateFixedSizeList(n: Int)` extension method on the `Generators.Implicits._`

jachro (approved):
Cool! Thanks Eike!

triples-generator/src/test/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/tooling/AllProjectsSpec.scala
eikek: Ah right thanks, sometimes I forget about these

PR #1223, opened 2022-12-06
feat: Project Access Token name to be configurable

jachro:
This PR introduces config value for the name of Project Access Tokens that are created on behalf of the user in GitLab.

It's to prevent different environments from using the same name tokens which could cause different environments to remove tokens from each other.


eikek (approved):
👍🏼

PR #1225, opened 2022-12-06
refactor: metrics to be passed through context bound

### LINKED ISSUE #913: Contextual objects to be passed implicitly
As a renku-graph developer, I'd like objects that have contextual meaning not to pollute the constructors' and factories' signatures and be passed implicitly using Context Bound.

Acceptance criteria (PRs to be created on the smallest reasonable chunks of changes):
- [ ] maybe `GitLabApiUrl`, `GitLabUrl`, `RenkuUrl`, `io.renku.config.renku.ApiUrl`;
- [x] `MetricsCollector`, like the `Histogram` and `Gauge`, not to be used as they are now as a single object type in different applications but a different subclass should be created per application. That way they are distinguishable from the type point of view and may be injected implicitly (e.g. there should be a `SentEventsGauge` type instead of the `SentEventsGauge` factory producing objects of the common `LabeledGauge` type). Single application `MetricsCollector` factories should be producing the same instance each time used (that's already achieved when a collector is registered in the `MetricsRegistry);
- [x] each application-specific type of the `MetricsCollector` might be newed up closer to the place when it's needed instead of the `Microservice` level; however, in terms of some collectors like `AwaitingGenerationGauge` the need of newing it up on the `Microservice` level might be kept because of the `GaugeResetScheduler` which takes them for periodic synchronisation - to be thought through as mentioned already the registration process returns the same instance anyway so having the gauge newed up on the `Microservice` can eventually leverage code readability or reasoning;
- [x] do the same as above for the `ExecutionTimeRecorder`;
- [x] do the same as above for the `QueriesExecutionTimes`;
- [x] `gitLabThrottler` to be passed implicitly wherever it's needed (as `Throttler` has two holes looking at the `io.renku.eventlog.SessionResource` might be a good hint);
- [x] `GitLabClient` to be newed up in the closest possible place it's used (alternatively, there might be a single instance of it and passed down to classes using it through the Context Bound);
- [ ] `RdfStoreConfig`
- [ ] add other cases if found;

jachro:
This PR makes the metrics objects be injected using context-bound

closes #913


event-log/src/main/scala/io/renku/eventlog/events/consumers/projectsync/SubscriptionFactory.scala
```diff
@@ -23,25 +23,25 @@ import cats.syntax.all._
...
 import org.typelevel.log4cats.Logger
 
 object SubscriptionFactory {
 
   def apply[F[
       _
-  ]: Async: NonEmptyParallel: Parallel: GitLabClient: AccessTokenFinder: MetricsRegistry: SessionResource: Logger](
-      queriesExecTimes: LabeledHistogram[F]
-  ): F[(EventHandler[F], SubscriptionMechanism[F])] = for {
+  ]: Async: NonEmptyParallel: Parallel: GitLabClient: AccessTokenFinder: SessionResource: Logger: MetricsRegistry: QueriesExecutionTimes]
```
eikek: I think Parallel extends NonEmptyParallel, so we could omit NonEmptyParallel from that list (but yeah, not much to gain :))

eikek (approved):
What an effort! 🚀 👍🏼

event-log/src/main/scala/io/renku/eventlog/events/consumers/projectsync/SubscriptionFactory.scala
jachro: Anyway, there's no effort to remove it.

PR #1234, opened 2022-12-08
feat: metrics to produce less data

### LINKED ISSUE #1219: KG services to produce less data the Grafana dashboards
As a renku maintainer, I'd like the KG Dashboards in Grafana to be more responsive.

Acceptance criteria:
- [ ] the SQL execution times metric to collect data only if a query took more than a value of a threshold
- [ ] the SPARQL execution times metric to collect data only if a query took more than a value of a threshold
- [ ] dashboards that show data from both the metrics to be renamed in a way it's obvious this is the slow queries exec times
- [ ] the metrics keeping number of events in certain statuses to keep info only about a certain number of projects, not all
  - [ ] fix the metrics so it doesn't go on minus if no value about a certain project is kept when updating to a negative value.

jachro:
This PR introduces numerous changes to the way we collect data for Prometheus.

closes #1219


graph-commons/src/main/scala/io/renku/metrics/Histogram.scala
```diff
@@ -47,18 +61,46 @@ class SingleValueHistogramImpl[F[_]: MonadThrow](val name: String Refined NonEmp
...
+
+    def observeDuration: F[Double] = MonadThrow[F].catchNonFatal {
+      timer.observeDuration()
+    }
+  }
+
+  final class ThresholdTimerImpl[F[_]: MonadThrow] private[metrics] (timer: LibHistogram.Timer,
+                                                                     labelValue:       String,
+                                                                     wrappedCollector: LibHistogram,
+                                                                     threshold:        Double
```
eikek: Maybe we can add some hints for what it is, like seconds, millis etc - maybe `thresholdMillis`? just a nitpick :)

graph-commons/src/main/scala/io/renku/metrics/LabeledGauge.scala
```diff
@@ -0,0 +1,176 @@
...
+    LibGauge
+      .build()
+      .name(name.value)
+      .help(help.value)
+      .labelNames(labelName.value)
+      .create()
+
+  private val zeroedValues = Ref.unsafe[F, List[(LabelValue, Instant)]](List.empty)
+
+  def clearZeroedValues(after: Duration = 2 minutes): F[Unit] = Spawn[F].start {
```
eikek: Do I understand correctly: this starts a background task which deletes 0-values in a fixed interval? Was thinking that maybe the background-starting behaviour could be indicated in the name (when reading `flatMap(clearZeroedValues)`  it might add more clarity)

eikek (approved):
Great effort! 🚀

graph-commons/src/main/scala/io/renku/metrics/LabeledGauge.scala
jachro: Good point! I'll change it to `startZeroedValuesCleaning`

eikek (commented):
Thank you!!

PR #1236, opened 2022-12-09
fix: deduplicate project created dates after transformation

eikek:
When events are processed concurrently it may happen that different project created dates get into the KG. With the "eventual consistency" requirements, we check for this situation after transformation and remove all but the earliest one.

/deploy #persist

### DIFF
--- a/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsprovisioning/transformation/namedgraphs/projects/UpdatesCreator.scala
+++ b/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsprovisioning/transformation/namedgraphs/projects/UpdatesCreator.scala
@@ -28,6 +28,7 @@ import io.renku.triplesstore.SparqlQuery.Prefixes
 
 private trait UpdatesCreator {
   def prepareUpdates(project:      Project, kgData: ProjectMutableData): List[SparqlQuery]
+  def postUpdates(project:         Project): List[SparqlQuery]
   def dateCreatedDeletion(project: Project, kgData: ProjectMutableData): List[SparqlQuery]
 }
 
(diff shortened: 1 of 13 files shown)


triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsprovisioning/transformation/namedgraphs/projects/UpdatesCreator.scala
```diff
@@ -170,4 +171,33 @@ private object UpdatesCreator extends UpdatesCreator {
...
+    SparqlQuery.of(
+      name = "transformation - project date created deduplicate",
+      Prefixes.of(schema -> "schema"),
+      s"""|WITH $resource 
+          |DELETE { ?p schema:dateCreated ?date }
+          |WHERE {
+          |  {
+          |    SELECT (min(?cdate) as ?minDate)
+          |    WHERE {
+          |      ?id a schema:Project;
```
jachro, line 190: It's a minor thing but perhaps selecting the `?id` in the subquery could allow you to use it in line 196 so you don't even have to check the type but select the date.

triples-generator/src/test/scala/io/renku/triplesgenerator/events/consumers/tsprovisioning/transformation/namedgraphs/projects/UpdatesCreatorSpec.scala
```diff
@@ -19,35 +19,125 @@
...
 
 import java.time.Instant
+import scala.concurrent.duration._
 
 class UpdatesCreatorSpec
     extends AnyWordSpec
     with IOSpec
     with should.Matchers
     with InMemoryJenaForSpec
+    // with ExternalJenaForSpec
```
jachro: I guess that's a leftover from debugging

PR #1237, opened 2022-12-12
fix: lucene query escape tool to treat phrases as strings

jachro:
This PR escapes search phrases passed to the Cross-Entity Search API so they are treated as strings.

The escape characters are taken from [here](https://lucene.apache.org/core/2_9_4/queryparsersyntax.html)

/deploy #persist

### DIFF
--- a/graph-commons/src/main/scala/io/renku/triplesstore/LuceneQueryEncoder.scala
+++ b/graph-commons/src/main/scala/io/renku/triplesstore/LuceneQueryEncoder.scala
@@ -18,37 +18,9 @@
 
 package io.renku.triplesstore
 
+import org.apache.lucene.queryparser.flexible.standard.QueryParserUtil
+
 object LuceneQueryEncoder {
 
-  private val specialCharactersEscape =
-    List("+",
-         "-",
-         "&",
-         "&&",
-         "|",
-         "||",
-         "!",
-         "(",
-         ")",
-         "{",
-         "}",
-         "[",
-         "]",
-         "^",
-         """"""",
-         "~",
-         "*",
-         "?",
-         ":",
-         """\""",
-         "/"
-    ).map {
-      case c if c == "}" || c == "]" || c == "*" || c == """\""" => c -> s"""\\$c"""
-      case c                                                     => c -> s"""\\\\$c"""
-    }.toMap
-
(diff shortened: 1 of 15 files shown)


graph-commons/src/main/scala/io/renku/triplesstore/LuceneQueryEncoder.scala
```diff
@@ -0,0 +1,54 @@
...
+         ":",
+         """\""",
+         "/"
+    ).map {
+      case c if c == "}" || c == "]" || c == "*" || c == """\""" => c -> s"""\\$c"""
+      case c                                                     => c -> s"""\\\\$c"""
+    }.toMap
+
+  def queryAsString(v: String): String =
+    specialCharactersEscape.foldLeft(v) { case (escaped, (specialChar, replacement)) =>
```
eikek: How does it work for strings like `"a&&b"` wrt the `&` vs `&&` replacements (same for `||`)? Wouldn't it be first replaced to `a\&\&b` so that the rule for `&&` would never trigger? We could perhaps use a list of tuples with sorting by rule preferences. Or it looks like we don't even need the `&&` and `||` rules? (then we could also use a `Char` type :))

eikek: OTOH maybe it worth thinking about using the java lucene library to let it escape these strings?

graph-commons/src/test/scala/io/renku/triplesstore/LuceneQueryEncoderSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+
+        val query = s"$specialChar${nonEmptyStrings(minLength = 3).generateOne}"
+        val name  = sentenceContaining(Refined.unsafeApply(query)).generateAs(projects.Name)
+        val project = renkuProjectEntities(visibilityPublic)
+          .modify(replaceProjectName(name))
+          .generateOne
+
+        upload(to = projectsDataset, project)
+
+        runSelect(
```
eikek, line 73: Since the query encoder is a pure function, maybe we can also add some simple "input -> expected output" tests? they should be much faster as well

eikek (approved):

PR #1238, opened 2022-12-12
fix: user's projects API to return all not activated projects

### LINKED ISSUE #1232: There's some weirdness going on when getting the user's projects
It looks like there's something weird going on when calling the User Projects endpoint in the case when the user has more than the requested `per_page` number of projects in GL that are not activated in KG.

We simply do not load all pages from GL.

jachro:
closes #1232

### DIFF
--- a/knowledge-graph/src/main/scala/io/renku/knowledgegraph/users/projects/finder/GLProjectFinder.scala
+++ b/knowledge-graph/src/main/scala/io/renku/knowledgegraph/users/projects/finder/GLProjectFinder.scala
@@ -24,6 +24,7 @@ import cats.effect.Async
 import cats.syntax.all._
 import io.renku.graph.model.{persons, projects}
 import io.renku.http.client.GitLabClient
+import io.renku.http.rest.paging.model.Page
 import io.renku.knowledgegraph.users.projects.model.Project
 import org.typelevel.log4cats.Logger
 
@@ -39,10 +40,12 @@ private object GLProjectFinder {
 private class GLProjectFinderImpl[F[_]: Async: GitLabClient: Logger](creatorFinder: GLCreatorFinder[F])
     extends GLProjectFinder[F] {
 
+  import GitLabClient.maybeNextPage
   import cats.syntax.all._
   import creatorFinder._
   import eu.timepit.refined.auto._
   import io.circe._
+  import io.renku.http.tinytypes.TinyTypeURIEncoder._
   import io.renku.tinytypes.json.TinyTypeDecoders._
   import org.http4s._
   import org.http4s.circe.jsonOf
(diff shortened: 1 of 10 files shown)


knowledge-graph/src/main/scala/io/renku/knowledgegraph/users/projects/finder/GLProjectFinder.scala
```diff
@@ -60,14 +63,20 @@ private class GLProjectFinderImpl[F[_]: Async: GitLabClient: Logger](creatorFind
...
-      mapResponse
-    )(criteria.maybeUser.map(_.accessToken))
+  private def findProjectsAndCreators(criteria: Criteria, page: Page = Page.first): F[List[ProjectAndCreator]] =
+    GitLabClient[F]
+      .get(uri"users" / criteria.userId / "projects" withQueryParam ("page", page), "user-projects")(mapResponse)(
+        criteria.maybeUser.map(_.accessToken)
+      )
+      .flatMap {
+        case (results, None)           => results.pure[F]
+        case (results, Some(nextPage)) => findProjectsAndCreators(criteria, nextPage).map(results ::: _)
```
eikek, line 73: We could also use a `fs2.Stream` to load it lazily (if we don't always need everything immediately). Might also help if GL should have some rate-limiting in place? Or should we handle this (better) explicitely?

knowledge-graph/src/main/scala/io/renku/knowledgegraph/users/projects/finder/ProjectsFinder.scala
```diff
@@ -51,7 +51,7 @@ private class ProjectsFinderImpl[F[_]: MonadThrow](
     (findProjectsInTS(criteria) -> findProjectsInGL(criteria))
       .mapN(mergeFavouringActivated)
       .map(filterBy(criteria.filters.state))
-      .map(_.sortBy(_.name))
+      .map(_.sortBy(_.name.value.toLowerCase))
```
eikek: We could also define the `Ordering` for this `name` type? (since we have specific types)

eikek (approved):

knowledge-graph/src/main/scala/io/renku/knowledgegraph/users/projects/finder/GLProjectFinder.scala
jachro: Yes, we could use streams here but actually, we need to fetch all the user projects as we then combine them with kg projects and only after we sort and page them on the Scala side. So I thought maybe it was not worth switching to Stream here.

knowledge-graph/src/main/scala/io/renku/knowledgegraph/users/projects/finder/ProjectsFinder.scala
jachro: Yes, good idea.

PR #1248, opened 2022-12-16
fix: Fix selecting workflows when modified ones exist

eikek:
When a fulltext search query is given, the entities that have a plan are selected, but without this query, the selection was missing. This resulted in not returning any plans.


jachro (approved):
👍🏻

PR #1256, opened 2022-12-20
feat: read avatar url from gitlabs project details

### LINKED ISSUE #1242: Add project image
As a graph API user, I'd like the Cross-Entity Search and Single Project APIs to return info about Project image.

The problem & solution
Our UI team would like to display the Project image for the user. Currently, only Datasets have images on KG. Before Project image(s) will be introduced on the CLI, we can use Project image that can be retrieved from the GL API.

Acceptance criteria:

- [x] adapt the transformation process
  - [x] read `avatar_url` returned by the [Single Project GL API](https://docs.gitlab.com/ee/api/projects.html#get-single-project) during event transformation
  - [x] encode the URL as an image URI similar to Datasets; we should extract the Image model from Dataset to some common space and share it with Project
  - [x] ensure the URL can be changed so the transformation process needs to update/delete it in the TS
- [x] adapt the Cross-Entity Search to return info about Project image and a direct link to it (see the Dataset images)
  - [x] update README
- [x] adapt the Single Project API to return info about Project image and a direct link to it (see the Dataset images)
  - [x] update README

Assumption:
* we won't add images to projects that are already in the TS (no re-provisioning)

eikek:
Read `avatar_url` from gitlabs single-project response.

Issue: #1242 

/deploy #persist

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -218,6 +218,8 @@ lazy val commonSettings = Seq(
     "-Ycache-macro-class-loader:last-modified", // and macro definitions. This can lead to performance improvements.
     "-Ywarn-value-discard" // Emit warning and location for usages of deprecated APIs.
   ),
+  Compile / console / scalacOptions := (Compile / scalacOptions).value.filterNot(_ == "-Xfatal-warnings"),
+  Test / console / scalacOptions := (Compile / console / scalacOptions).value,
   // Format: on
   organizationName := "Swiss Data Science Center (SDSC)",
   startYear := Some(java.time.LocalDate.now().getYear),
--- a/entities-search/src/test/scala/io/renku/entities/search/WorkflowsEntitiesFinderSpec.scala
+++ b/entities-search/src/test/scala/io/renku/entities/search/WorkflowsEntitiesFinderSpec.scala
@@ -21,6 +21,7 @@ package io.renku.entities.search
 import Criteria._
 import EntityConverters._
 import cats.syntax.all._
+import io.renku.entities.search
 import io.renku.entities.search.model.Entity.Workflow.WorkflowType
 import io.renku.generators.Generators.Implicits._
 import io.renku.generators.Generators._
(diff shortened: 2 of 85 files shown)


build.sbt
```diff
@@ -218,6 +218,8 @@ lazy val commonSettings = Seq(
     "-Ycache-macro-class-loader:last-modified", // and macro definitions. This can lead to performance improvements.
     "-Ywarn-value-discard" // Emit warning and location for usages of deprecated APIs.
   ),
+  Compile / console / scalacOptions := (Compile / scalacOptions).value.filterNot(_ == "-Xfatal-warnings"),
```
jachro, line 221: I guess we'd like to take the filter out at the end.

entities-search/src/test/scala/io/renku/entities/search/WorkflowsEntitiesFinderSpec.scala
```diff
@@ -184,4 +185,19 @@ class WorkflowsEntitiesFinderSpec
       wfs.map(_.workflowType) should contain theSameElementsAs List(WorkflowType.Step, WorkflowType.Composite)
     }
   }
+
```
jachro: Shouldn't the new test go to a new `ProjectEntitiesFinderSpec`?

build.sbt
eikek: This only applies when you run `console` - compile and test scalac settings are not changed.

entities-search/src/test/scala/io/renku/entities/search/WorkflowsEntitiesFinderSpec.scala
eikek: Oh, right 🤦🏼 - I think I wanted it in `EntityFindersSpec`

jachro (approved):
The changes look really good! It's been quite a bit chunk of work. Thanks Eike!

entities-search/src/test/scala/io/renku/entities/search/WorkflowsEntitiesFinderSpec.scala
jachro: Yes, that's also good.

PR #1258, opened 2022-12-22
refactor: a new entities-search module to deal with searching

### LINKED ISSUE #1257: KG API to look for Datasets in the Datasets graph
As a KG API user, I'd like searching for Datasets is fast.

**The problem**
Currently, finding a Dataset in the TS is difficult and slow. There are a number of reasons for that, where the most significant are:
* huge amount of data due to the way Datasets are modelled in the TS;
* no single place where the data can be found; the search has to go through all Project graphs

**The idea:**
It looks like creating a dedicated Datasets graph where only findable/discoverable Datasets should be kept might be a good solution. Such a graph would keep info about Datasets that are leaves of either the import or the derivation graphs for each Dataset. Additionally, each Dataset in the proposed graph would need to keep a link to a Project (or Projects) where it belongs so, in case more information is needed, a quick jump to the graph with full Dataset info can be done. It's worth mentioning, the new Datasets graph would not contain the full info about each Dataset but only the ones needed for the search queries.

**Acceptance criteria:**

- [x] transformation flow changes
  - [x] what we need to keep in the Datasets graph (for each DS)?
    - [x] resourceId, name, visibility (the most visible), date, creators, keywords, description, images, links to projectIds
  - [x] how we compute the data?
    - [x] for each DS in the payload, collect the last modification (except the last modification is invalidation)
    - [x] all invalidated DS should be removed from the new graph (either remove the DS from the graph or just a link to a project if there are more than one)
    - [x] the change to the new graph needs to be done in an upsert way
  - [x] how we put the data to the graph?
    - [x] create a new module with read and write APIs; both TG and KG to use the API
- [x] how do we put the data for all the Datasets we've got in the TS?
- [x] Cross-Enitity Search API changes
- [ ] Make sure the UI can work with the new way Cross-Enitity Search returns info/links about Datasets (topmostSameAs instead of identifier)
- [ ] ~Free-Text DS Seach API changes~ - we don't have to do it as this API will be removed in the future.

resolves #1077

jachro:
This PR introduces a new `entities-search` module. For the time being it contains the finder for the Cross-Entity search API.


eikek (approved):
🚀

PR #1261, opened 2022-12-22
feat: Datasts graph provisioning process

### LINKED ISSUE #1257: KG API to look for Datasets in the Datasets graph

jachro:
This PR adds a new process to provision a new `Datasets` graph in the TS. The new graph will contain only Datasets that should be discoverable for the search APIs. Additionally, it will contain only a subset of Dataset entity properties.

/deploy #persist


build.sbt
```diff
@@ -39,13 +58,25 @@ lazy val generators = Project(
...
+)
+
 lazy val tinyTypes = Project(
   id = "tiny-types",
   base = file("tiny-types")
 ).settings(
   commonSettings
 ).dependsOn(
-  generators % "test->test"
+  triplesStoreClient % "compile->compile",
```
eikek: sbt can combine the scope mappings in a single string like `% "compile->compile; test->test"` - but totally fine like this as well, ofc

entities-search/src/test/scala/io/renku/entities/searchgraphs/commands/QuadsEncodersSpec.scala
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.renku.entities.searchgraphs.commands
+
+import org.scalatest.matchers.should
+import org.scalatest.wordspec.AnyWordSpec
+
+class QuadsEncodersSpec extends AnyWordSpec with should.Matchers {}
```
eikek: Is this intentionally empty?

renku-model/src/main/scala/io/renku/graph/model/entities/Dataset.scala
```diff
@@ -499,41 +499,64 @@ object Dataset {
...
-        DataProperty(renku / "originalIdentifier", xsd / "string"),
-        DataProperty(prov / "invalidatedAtTime", xsd / "dateTime"),
-        DataProperty(schema / "description", xsd / "string"),
-        DataProperty(schema / "keywords", xsd / "string"),
-        DataProperty(schema / "license", xsd / "string"),
-        DataProperty(schema / "version", xsd / "string")
-      ),
-      ReverseProperties(PublicationEvent.ontology)
-    )
+  object Ontology {
```
eikek, line 502: ❤️

renku-model-tiny-types/src/main/scala/io/renku/graph/model/images/Image.scala
```diff
@@ -54,9 +54,16 @@ object Image {
     } yield Image(resourceId, uri, position)
   }
 
-  val ontology: Type = Type.Def(
-    Class(schema / "ImageObject"),
-    DataProperty(schema / "contentUrl", xsd / "string"),
-    DataProperty(schema / "position", xsd / "int")
-  )
+  object Ontology {
```
eikek, line 57: Thanks!

eikek (approved):
Very nice! 🚀

build.sbt
jachro: Yes, I didn't know you can do it like this but saw it on your PR. I'll change it here.

entities-search/src/test/scala/io/renku/entities/searchgraphs/commands/QuadsEncodersSpec.scala
jachro: This bit PRs... I simply forgot about adding tests :)

jachro: Wrong, actually, not the tests are in the `EncodersSpec`

PR #1268, opened 2023-01-04
chore: Replace compatibility-matrix with a simpler config

### LINKED ISSUE #1259: Add support for schema V10

**Acceptance criteria:**

- [ ] immediate changes
  ```
  ```

eikek:
The compatibility-matrix was used to specify which cli-version was compatible with a schema-version. Based on this a decision whether to trigger re-provisioning was derived. Now the config only contains the expected CLI version, the expected schema version and whether to kick off re-provisioning if, and only if, cli or schema version is different from the last run.


triples-generator/src/test/scala/io/renku/triplesgenerator/config/VersionCompatibilityConfigSpec.scala
```diff
@@ -39,88 +37,63 @@ class VersionCompatibilityConfigSpec extends AnyWordSpec with should.Matchers {
...
-      val config = ConfigFactory.parseMap(
-        Map("compatibility-matrix" -> unparsedConfigElements).asJava
-      )
+      val config = ConfigFactory.parseString(unparsedConfigElements)
 
-      readVersionPairs(config) shouldBe NonEmptyList.fromListUnsafe(expected).pure[Try]
+      readCompatibilityConfig(config) shouldBe c.pure[Try]
     }
 
     "return a list of RenkuVersionPairs with the first element of the RenkuDevVersion if it exists and log a warning" in new TestCase {
```
jachro: Maybe we could adapt the name of the test so it doesn't mention the VersionPairs but the new CompatibilityConfig

triples-generator/src/test/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/reprovisioning/ReProvisionJudgeSpec.scala
```diff
@@ -42,59 +41,32 @@ class ReProvisionJudgeSpec extends AnyWordSpec with should.Matchers with MockFac
     "return true when TS schema version cannot be found" in new TestCase {
       (versionPairFinder.find _).expects().returning(Option.empty[RenkuVersionPair].pure[Try])
 
-      judge(NonEmptyList.one(renkuVersionPairs.generateOne)).reProvisioningNeeded() shouldBe true.pure[Try]
+      judge(compatibilityGen.generateOne).reProvisioningNeeded() shouldBe true.pure[Try]
     }
 
     "return true when TS schema version is different than the matrix schema version" in new TestCase {
```
jachro: Maybe the 'config' word should be used instead of the 'matrix' in the test name. Perhaps, we can also mention that the `reProvisioningNeeded` has to be set to `true`

triples-generator/src/test/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/reprovisioning/ReProvisionJudgeSpec.scala
```diff
@@ -42,59 +41,32 @@ class ReProvisionJudgeSpec extends AnyWordSpec with should.Matchers with MockFac
...
-      val matrixVersionPairs = renkuVersionPairs.generateNonEmptyList(min = 2)
+      val configVersionPairs = compatibilityGen
+        .suchThat(_.schemaVersion != tsVersionPair.schemaVersion)
+        .map(_.copy(reProvisioningNeeded = true))
 
-      judge(matrixVersionPairs).reProvisioningNeeded() shouldBe true.pure[Try]
+      judge(configVersionPairs.generateOne).reProvisioningNeeded() shouldBe true.pure[Try]
     }
 
-    "return true when the top two schema versions in the matrix are the same " +
```
jachro, line 57: Maybe it would be also good to add a test (I couldn't find one) where will be expecting `false` when versions match but the `reProvisioningNeeded` is `true`.

jachro: I think we also miss a test that checks the judge to return `false` for different `cliVersion`s and/or `schemaVersion`s and the `reProvisioningNeeded=false`

triples-generator/src/test/scala/io/renku/triplesgenerator/init/CliVersionCompatibilityVerifierImplSpec.scala
```diff
@@ -31,25 +30,21 @@ import scala.util.{Failure, Success, Try}
 class CliVersionCompatibilityVerifierImplSpec extends AnyWordSpec with MockFactory with should.Matchers {
 
   "run" should {
-    "return Unit if the cli version matches the cli version from the first pair in compatibility matrix" in {
-      val cliVersion = cliVersions.generateOne
-      val versionPairs = renkuVersionPairs.generateNonEmptyList() match {
-        case NonEmptyList(head, tail) => NonEmptyList(head.copy(cliVersion = cliVersion), tail)
-      }
-      val checker = new CliVersionCompatibilityVerifierImpl[Try](cliVersion, versionPairs)
+    "return Unit if the cli version matches the cli version from the first pair in compatibility config" in {
```
jachro: I think the `first pair` in the presence of the new compatibility config does not hold anymore.

jachro (approved):

triples-generator/src/test/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/reprovisioning/ReProvisionJudgeSpec.scala
eikek: I think we have the last one at least covered in some of the tests that also mention `underReprovisioning` - but it's hard to see and I added these two cases explicitely.

PR #1272, opened 2023-01-05
chore: Ensure plans are decoded using strict type annotations

### LINKED ISSUE #1259: Add support for schema V10

eikek:
This avoids accidentally parsing sub-types that are not supported for now.

Issue: #1259

### DIFF
--- a/renku-model/src/main/scala/io/renku/graph/model/entities/Plan.scala
+++ b/renku-model/src/main/scala/io/renku/graph/model/entities/Plan.scala
@@ -61,9 +61,22 @@ object Plan {
     }
 
   implicit def decoder(implicit renkuUrl: RenkuUrl): JsonLDDecoder[Plan] =
-    CompositePlan.decoder
-      .asInstanceOf[JsonLDEntityDecoder[Plan]]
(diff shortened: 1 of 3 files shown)


renku-model/src/main/scala/io/renku/graph/model/entities/Plan.scala
```diff
@@ -257,8 +270,11 @@ object StepPlan {
         )
     }
 
+  private val withStrictEntityTypes: Cursor => JsonLDDecoder.Result[Boolean] =
+    _.getEntityTypes.map(types => types.toList.toSet == StepPlan.entityTypes.toList.toSet)
```
jachro: The `EntityTypes` class has a custom `equals` method so you should be able to do:
```
....map(_ == StepPlan.entityTypes)

jachro (approved):

renku-model/src/main/scala/io/renku/graph/model/entities/Plan.scala
eikek: Ah 🤦🏼 Thank you!

PR #1274, opened 2023-01-06
chore: Map dateModified to dateCreated

### LINKED ISSUE #1259: Add support for schema V10

eikek:
When CLI passes a dateModified, it copies dateCreated from the first dataset or plan. We keep dateCreated on each entity as the date the specific entity was created and so map the dateModified to our dateCreated. To get the dateCreated of the original dataset/plan, we can follow the `topMostSameAs` edge.

Issue: #1259


jachro (approved):
Looks good in my eyes

renku-cli-model/src/main/scala/io/renku/cli/model/CliModel.scala
```diff
@@ -21,7 +21,7 @@ package io.renku.cli.model
 import io.renku.jsonld.{JsonLD, JsonLDEncoder}
 
 /** Marker trait to annotate a class that models an entity from the cli schema. */
-trait CliModel
+trait CliModel extends Product with Serializable
```
eikek: Is this really necessary? i.e. the `Serializable`?

renku-cli-model/src/test/scala/io/renku/cli/model/diffx/CliDiffInstances.scala
```diff
@@ -24,71 +24,77 @@ import io.renku.graph.model.diffx.ModelTinyTypesDiffInstances
 
 trait CliDiffInstances extends ModelTinyTypesDiffInstances {
 
-  implicit val cliPersonDiff: Diff[CliPerson] = Diff.derived[CliPerson]
+  implicit lazy val cliPersonDiff: Diff[CliPerson] = Diff.derived[CliPerson]
```
eikek, line 27: I tried to get the order right: :-) but `lazy` is actually more convenient 👍🏼

renku-cli-model/src/test/scala/io/renku/cli/model/generators/SoftwareAgentGenerators.scala
```diff
@@ -0,0 +1,34 @@
...
+package io.renku.cli.model.generators
+
+import io.renku.cli.model.CliSoftwareAgent
+import io.renku.graph.model.RenkuTinyTypeGenerators
+import org.scalacheck.Gen
+
+trait SoftwareAgentGenerators {
+
+  def softwareAgentGen: Gen[CliSoftwareAgent] =
+    for {
```
eikek, line 28: Just wondering: I saw you changed many for comprehensions to start on the `def` line - but this one is on the line below (which is my personal pref 😆 so I like it hehe). Is this perhaps a written formatting rule somewhere?

eikek (commented):
Epic!! 🚀

renku-cli-model/src/main/scala/io/renku/cli/model/CliModel.scala
jachro: Actually, I think I can take it out

renku-cli-model/src/test/scala/io/renku/cli/model/diffx/CliDiffInstances.scala
jachro: You caught me on naughty things :)

renku-cli-model/src/test/scala/io/renku/cli/model/generators/SoftwareAgentGenerators.scala
jachro: I don't think we've got a rule rewriting that anywhere. I somehow have a clashing preference so it might be that when I do some changes in a file where I see your formatting, my auto-manual formatter could kick-off. Sorry...

eikek: No worries! I don't have a strong opinion here and might also mix it sometimes. was genuinely wondering as I looked through

PR #1292, opened 2023-01-18
feat: TS migration kicking-off background migration to schema V10

jachro:
This PR brings a migration that will effectively re-provision all the Projects found in the TS. The process doesn't wipe out the TS but rather redoes each Project one by one. As a result, API users will be able to find the relevant data throughout the duration of the operation. At a single point in time, a maximum of (number of TGs running on an environment) Projects will be unavailable.

### DIFF
--- a/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToV10.scala
+++ b/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToV10.scala
@@ -33,22 +33,21 @@ import io.renku.metrics.MetricsRegistry
 import io.renku.triplesstore.SparqlQuery.Prefixes
 import io.renku.triplesstore._
 import org.typelevel.log4cats.Logger
-import tooling.{RecordsFinder, RecoverableErrorsRecovery}
+import tooling._
 
-private class MigrationToV10[F[_]: Async](
-    projectsFinder:   ProjectsFinder[F],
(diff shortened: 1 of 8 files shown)


triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToV10.scala
```diff
@@ -0,0 +1,127 @@
...
+import io.renku.triplesstore.SparqlQuery.Prefixes
+import io.renku.triplesstore._
+import org.typelevel.log4cats.Logger
+import tooling.{RecordsFinder, RecoverableErrorsRecovery}
+
+private class MigrationToV10[F[_]: Async](
+    projectsFinder:   ProjectsFinder[F],
+    eventsSender:     EventSender[F],
+    recoveryStrategy: RecoverableErrorsRecovery = RecoverableErrorsRecovery
+) extends Migration[F] {
```
eikek: I'm wondering if it should be a "registered" migration? It should only run once - or perhaps this is addressed already and I didn't see it

eikek (approved):
👍🏼

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/MigrationToV10.scala
jachro: I think you are right. I think I missed that bit. Good catch

PR #1295, opened 2023-01-19
chore: bump rp to 1.11.0

rokroskar:
bumps RP to 1.11.0

/deploy


jachro:
I guess we don't need this one anymore

PR #1296, opened 2023-01-19
fix: Plan entity decoder not to be strict on types

### LINKED ISSUE #1259: Add support for schema V10

jachro:
This PR takes out the strict type check predicate from both the Plan entity decoders. The reason is that introduction of `WorkflowFile` version of both the Plans does not bring significant changes to any of the non-`WorkflowFile` Plans. Moreover, not decoding `WorkflowFile` Plans causes failures during decoding Activities created as a result of `WorkflowFile` Plan execution. The option here would be to skip decoding `WorkflowFile` related Activities but this seems not to be a trivial thing to do.
I tried decoding a payload from a project using WorkflowFile feature and without the strict type check on Plan decoders all seems to be working fine.

### DIFF
--- a/renku-model/src/main/scala/io/renku/graph/model/entities/Plan.scala
+++ b/renku-model/src/main/scala/io/renku/graph/model/entities/Plan.scala
@@ -19,8 +19,8 @@
 package io.renku.graph.model
 package entities
 
-import StepPlanCommandParameter.{CommandInput, CommandOutput, CommandParameter}
 import Schemas.{prov, renku, schema}
+import StepPlanCommandParameter.{CommandInput, CommandOutput, CommandParameter}
 import cats.data.{NonEmptyList, Validated, ValidatedNel}
 import cats.syntax.all._
 import io.circe.DecodingFailure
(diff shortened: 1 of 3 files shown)


renku-model/src/test/scala/io/renku/graph/model/entities/PlanSpec.scala
```diff
@@ -186,45 +187,15 @@ class PlanSpec
...
-      val newJson =
-        JsonLDTools
-          .view(sp)
-          .selectByTypes(entities.StepPlan.entityTypes)
-          .addType(renku / "WorkflowPlan")
-          .value
+      val decoded = jsonLD.cursor.as[List[entities.Plan]]
 
-      newJson.cursor.as[List[entities.StepPlan]] shouldBe Right(Nil)
+      decoded.fold(err => fail(err.message), identity) should contain theSameElementsAs List(sp, cp)
```
eikek, line 198: A test for the case we have an additional type in the jsonld would also be useful, no? like the removed `fail decode if…` tests (that didn't really fail, but returned an empty list) but on the `Plan.decoder` level (not on the subtypes) - iiuc that was the problem yesterday?

renku-model/src/main/scala/io/renku/graph/model/entities/Plan.scala
```diff
@@ -61,21 +61,15 @@ object Plan {
...
-        }
+    JsonLDDecoder.entity(EntityTypes of prov / "Plan") { cursor =>
+      lazy val noMatchFailure: Either[DecodingFailure, Plan] = cursor.getEntityTypes.flatMap { ets =>
+        DecodingFailure(
+          DecodingFailure.Reason.CustomReason(show"Cannot decode entity as Plan: $ets"),
+          cursor.jsonLD.toJson.hcursor
+        ).asLeft
+      }
+
+      StepPlan.decoder.apply(cursor) orElse CompositePlan.decoder.apply(cursor) orElse noMatchFailure
```
eikek, line 72: Interesting! I _think_ I tried a similar solution at first but it failed some tests I created 🤔  (maybe the tests were not good 😄)

jachro: Hm, don't really know what to say

renku-model/src/test/scala/io/renku/graph/model/entities/PlanSpec.scala
jachro: Yes, we could have one. Let me think if I can compile something like that.

renku-model/src/main/scala/io/renku/graph/model/entities/Plan.scala
eikek: Sorry, it was just a thought in my head, nothing to answer :)

eikek (approved):

PR #1301, opened 2023-01-20
chore: Cli model module

eikek:
This PR introduces new `cli-model-module` containing all model classes matching schema definitions on the renku-python project.


eikek:
Thanks!! 🎉

PR #1303, opened 2023-01-24
feat: Allow multiple sort parameters

### LINKED ISSUE #1278: Cross-Entity Search to take multiple values for the sort param
As a graph API user, I'd like to be able to sort by multiple fields.

Acceptance criteria:
- [ ] the `sort` parameter to allow multiple values so we can ask the TS to perform the requested order

eikek:
This adds support for specifying multiple `sort` query parameters for the cross entity search.

There is a new type introduced to capture multiple sort parameters (based on a non-empty list). It is `Sorting` in the rest utility package and `OrderBy` for the triples store client. To lower confusion about names, the subtypes of `rest.SortBy`that were named `Sorting`are renamed to `Sort`.

The change is also applied to the find events endpoint, because it uses most of the same machinery and I kind of converted it automatically :).

closes #1278


jachro (approved):
Great work Eike! Thanks!

PR #1317, opened 2023-02-03
feat: creator names only for the requested page of User's Projects

### LINKED ISSUE #1316: Call to user's projects timeouts
It looks like for a user owning a bigger number of projects, the https://renku-kg-dev.dev.renku.ch/knowledge-graph/users/12193/projects?state=not_activated timeouts

jachro:
This PR improves the algorithm that finds the results for the User's Projects API. The new version finds the creators' names only for the page of results that is returned to the caller.

closes #1316 
/deploy #persist


eikek (approved):
👍🏼

PR #1319, opened 2023-02-07
fix: optional model String TTs to be decoded to None when blank

### LINKED ISSUE #1259: Add support for schema V10

jachro:
This PR fixes the model String TTs so that they get decoded to None in case of a blank value instead of failing.

### DIFF
--- a/knowledge-graph/src/test/scala/io/renku/knowledgegraph/projects/details/ProjectJsonLDEncoderSpec.scala
+++ b/knowledge-graph/src/test/scala/io/renku/knowledgegraph/projects/details/ProjectJsonLDEncoderSpec.scala
@@ -49,14 +49,14 @@ class ProjectJsonLDEncoderSpec extends AnyWordSpec with should.Matchers with Sca
         resourceId   <- cursor.downEntityId.as[projects.ResourceId]
         identifier   <- cursor.downField(schema / "identifier").as[projects.GitLabId]
         path         <- cursor.downField(renku / "projectPath").as[projects.Path]
-        name         <- cursor.downField(schema / "name").as[projects.Name]
+        name         <- cursor.downField(schema / "name").as[Option[projects.Name]].flatMap(projects.Name.failIfNone)
         maybeDesc    <- cursor.downField(schema / "description").as[Option[projects.Description]]
         visibility   <- cursor.downField(renku / "projectVisibility").as[projects.Visibility]
         dateCreated  <- cursor.downField(schema / "dateCreated").as[projects.DateCreated]
         maybeCreator <- cursor.downField(schema / "creator").as[Option[Creator]]
         dateUpdated  <- cursor.downField(schema / "dateModified").as[DateUpdated]
         maybeParent  <- cursor.downField(prov / "wasDerivedFrom").as[Option[ParentProject]]
-        keywords     <- cursor.downField(schema / "keywords").as[List[projects.Keyword]]
+        keywords     <- cursor.downField(schema / "keywords").as[List[Option[projects.Keyword]]].map(_.flatten)
         maybeVersion <- cursor.downField(schema / "schemaVersion").as[Option[SchemaVersion]]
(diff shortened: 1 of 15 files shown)


renku-model-tiny-types/src/main/scala/io/renku/graph/model/views/NonBlankTTJsonLDOps.scala
```diff
@@ -0,0 +1,59 @@
...
+
+trait NonBlankTTJsonLDOps[TT <: StringTinyType] extends JsonLDTinyTypeEncoder[TT] with StringTinyTypeJsonLDDecoders {
+  self: TinyTypeFactory[TT] with NonBlank[TT] =>
+
+  implicit lazy val optionalTTDecoder: JsonLDDecoder[Option[TT]] =
+    decodeBlankStringToNone[TT](self)
+
+  val failIfNone: Option[TT] => JsonLDDecoder.Result[TT] = {
+    case Some(v) => v.asRight
+    case None    => DecodingFailure(s"A value of '$typeName' expected but got any", Nil).asLeft
```
eikek: Not sure about the wording: ".. got any" - maybe ".. got none"? or "none or empty"?

knowledge-graph/src/test/scala/io/renku/knowledgegraph/projects/details/ProjectJsonLDEncoderSpec.scala
```diff
@@ -49,14 +49,14 @@ class ProjectJsonLDEncoderSpec extends AnyWordSpec with should.Matchers with Sca
         resourceId   <- cursor.downEntityId.as[projects.ResourceId]
         identifier   <- cursor.downField(schema / "identifier").as[projects.GitLabId]
         path         <- cursor.downField(renku / "projectPath").as[projects.Path]
-        name         <- cursor.downField(schema / "name").as[projects.Name]
+        name         <- cursor.downField(schema / "name").as[Option[projects.Name]].flatMap(projects.Name.failIfNone)
```
eikek, line 52: Not sure if I understand: if we do `failIfNone` at the end - wouldn't it make sense to use the previous decoder, that fails by itself on a blank value? (due to the NonBlank constraint)?

eikek (approved):
Thanks!!

renku-model-tiny-types/src/main/scala/io/renku/graph/model/views/NonBlankTTJsonLDOps.scala
jachro: Yes, I was puzzled here, too. I'll try to come up with something better :)

knowledge-graph/src/test/scala/io/renku/knowledgegraph/projects/details/ProjectJsonLDEncoderSpec.scala
jachro: Hmm, that's a good question. I somehow thought the previous decoder wouldn't be needed anymore due to the fact we know these fields may be coming with blank values. We could leave the old decoders along with the new ones producing Options, however, we might go into potential issues where we decode to Option as there will be a bit of ambiguity for the compiler. You know what I mean, the compiler will need to choose between
* the new decoder producing Option[TT]
* or the old decoder producing TT combined with the standard optionalDecoder.
I somehow find myself leaning towards not keeping the old decoder on these few types and using the `failIfNone` eventually if those few places.
Any more thoughts?

eikek: Ah I see, I somehow missed that the old ones were removed. I was hoping the compiler is fine if certain types only provide decoders for `Option[TT]` and others do for `T`.  But you're right, it could be a problem! I'm anyways totally fine wtih the final flatMap in a few cases

PR #1330, opened 2023-02-13
fix: SearchInfoFeatcher to work with ',' in names

jachro:
Apparently, there are `,`s appearing names of Person entities. Unfortunately, the `SearchInfoFeatcher` was using that char to separate rows in the results set.
This PR fixes that.


entities-search/src/main/scala/io/renku/entities/searchgraphs/commands/SearchInfoFetcher.scala
```diff
@@ -106,14 +106,16 @@ private class SearchInfoFetcherImpl[F[_]: Async: Logger: SparqlQueryTimeRecorder
         |""".stripMargin
   )
 
+  private lazy val rowsSeparator = "##"
```
eikek: Maybe we could use the null byte? But not sure if jena/sparql supports this. This would be even harder to type in for users 😄 Just thinking when the first name with a double hash occurs 😰

eikek (approved):

entities-search/src/main/scala/io/renku/entities/searchgraphs/commands/SearchInfoFetcher.scala
jachro: Yes, that's definitely a stronger choice 👍🏻

PR #1331, opened 2023-02-13
chore: paged response calculation error

jachro:
This PR fixes an error where PagedResponse was failing when the requested page was beyond the max page having any results.


eikek (approved):
👍🏼

PR #1332, opened 2023-02-13
feat: EventSender to take service url on instantiation

jachro:
This PR makes the `EventSender` take service URL factory instead of using hardcoded `EventLogUrl`

### DIFF
--- a/graph-commons/src/main/scala/io/renku/events/producers/EventSender.scala
+++ b/graph-commons/src/main/scala/io/renku/events/producers/EventSender.scala
@@ -18,23 +18,24 @@
 
 package io.renku.events.producers
 
+import cats.{Applicative, Eval}
 import cats.effect.{Async, Temporal}
 import cats.syntax.all._
-import cats.{Applicative, Eval}
 import eu.timepit.refined.api.Refined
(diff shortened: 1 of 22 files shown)


graph-commons/src/main/scala/io/renku/events/producers/EventSender.scala
```diff
@@ -48,8 +49,19 @@ trait EventSender[F[_]] {
...
+    serviceUrl      <- consumerUrl()
+    sentEventsGauge <- SentEventsGauge[F]
+  } yield new EventSenderImpl(serviceUrl, sentEventsGauge, onErrorSleep = 15 seconds)
+
+  final case class EventContext(categoryName: CategoryName, errorMessage: String)
+}
+
 class EventSenderImpl[F[_]: Async: Logger](
-    eventLogUrl:            EventLogUrl,
+    serviceUrl:             UrlTinyType,
```
eikek: I was wondering why not use the new trait `EventConsumerUrl` as argument type? Can't see it used anywhere (but I probably missed it).

eikek (approved):

Issue #1168, opened 2022-11-03
Move jsonld decoders into triple-generator and indicate its purpose


eikek:
The JsonLD decoders are used for decoding data coming from the CLI. The encoders however are used to provide the representation handed to Jena. Often both, decoders and encoders are next to each other indicating some kind of matching (like that decoders could decode the output of the encoders) - but that is not necessarily true.

Therefore it would be beneficial for readability to move decoders into a separate namespace that indicates its purpose - decoding CLI data. Since this is only used in the triple generator, it makes sense to have it there.

The plan:
- [x] Create conversion from test entities to cli entities
- [x] Change tests to use these conversions to create cli entities
- [x] Change prod. decoders to go through the cli entities
- [x] Move decoders into proper module (tg)

PR #1336, opened 2023-02-14
feat: Decoders to use cli model

eikek:
The decoders for the model entities are now going through the cli data model and don't define their own decoders anymore.

On the model entities a new method `fromCli` is added to the companion objects. It accepts the corresponding cli data (and potentially some more supporting data) and creates a `ValidatedNel[String, A]` as e result. For consistency it always returns a `ValidatedNel` even when sometimes it wouldn't be necessary.

Issue: #1168


jachro (approved):
Cool! That's a pretty big chunk of work!

PR #1345, opened 2023-02-17
feat: v10 migration to be backpressured by env provisioning capabilities

### LINKED ISSUE #1338: V10 migration to produce events only if there are free resources
It was identified that the current version of the V10 migration would cause data outage (partial) for almost the whole time when projects are redone. This might be at least two weeks which is far from perfect or acceptable. The other important fact is that such a situation is not essential from the perspective of the migration process.

The solution is to make the migration be back pressured by the processing capabilities of the environment. More specifically, the migration should not generate any new event if all the subscribers to the `AWAITING_GENERATION` events are busy.

Acceptance criteria:

- [x] new status endpoint on the EL
- [ ] the V10 migration to be run async
- [ ] the V10 migration to store a `running` flag in the TS so it can be re-triggered when died/cancelled before completion
- [ ] the V10 migration to wait until there are free resources in the env before issuing a new event
- [ ] how to prevent no duplicate events are sent in a case of migration restart? Should we store them in the TS and remove only when events are sent for all the projects?

jachro:
This PR changes the v10 migration according to the points mentioned in #1338


eikek (approved):
👍🏼

PR #1346, opened 2023-02-17
chore: remove decoders from model entities

### LINKED ISSUE #1168: Move jsonld decoders into triple-generator and indicate its purpose

eikek:
The production model entities' decoders are now fully moved to the cli module. They can be removed now, while tests are adopted from converting from cli data into production data. The only decoder for decoding cli payloads is now in the triple-generator module. While adopting tests, some cases have been moved to the cli module.

Issue: #1168


jachro (approved):
Wow! Amazing! Things look so much cleaner now!

PR #1349, opened 2023-02-21
refactor: same subscription types to be used for receiving and sending

jachro:
This PR brings some refactoring work around sending and receiving event subscriptions by making the same types being shared on both sides.


event-log/src/main/scala/io/renku/eventlog/events/producers/Subscribers.scala
```diff
@@ -18,15 +18,15 @@
...
-import cats.{Applicative, MonadThrow, Show}
-import io.renku.events.CategoryName
-import io.renku.events.consumers.subscriptions.SubscriberUrl
+import io.renku.events.{CategoryName, Subscription}
+import io.renku.events.Subscription.SubscriberUrl
 import org.typelevel.log4cats.Logger
 
-private trait Subscribers[F[_], -SI <: SubscriptionInfo] {
-  def add(subscriptionInfo: SI): F[Unit]
+private trait Subscribers[F[_], -S <: Subscription.Subscriber] {
```
eikek, line 28: I was wondering if its necessary to define the `S` type parameter, I only saw implementations working with the abstract type so far. BUT: I can easily miss lots of things from looking at it in github, so it's meant as a genuine question. Reason is only in maybe simplifying the signatures a bit :)

eikek (approved):
🚀 great work!!

event-log/src/main/scala/io/renku/eventlog/events/producers/Subscribers.scala
jachro: Yes, I wanted to take that out as well. Unfortunately, there's one special Subscriber type for the TS Migration Request events which as far as I remember made this additional type necessary.

PR #1353, opened 2023-02-22
chore: Move from http4s blaze backend to ember

### LINKED ISSUE #1329: Migrate from blaze to ember
As a graph maintainer, I'd like graph services to use Ember instead of Blaze.

The reason is that Ember supposes to be a cleaner, newer and faster implementation. A bit more can be found [here](https://www.reddit.com/r/scala/comments/nu5iuy/http4s_blaze_vs_ember/)

eikek:
This PR changes the http4s backend from blaze to ember, which is the newer and better maintained http4s backend recommended to use by http4s.

It also contains a tiny refactoring removing parens `()` from a non-sideeffect method that carries a bit of adaptions 😅.

Closes: #1329

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -12,8 +12,9 @@ object Dependencies {
     val circeOptics            = "0.14.1"
     val diffx                  = "0.8.2"
     val http4s                 = "0.23.18"
-    val http4sBlaze            = "0.23.13"
+    val http4sBlaze            = "0.23.18"
     val http4sPrometheus       = "0.24.3"
+    val ip4s                   = "3.2.0"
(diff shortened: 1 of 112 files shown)


project/Dependencies.scala
```diff
@@ -12,8 +12,9 @@ object Dependencies {
     val circeOptics            = "0.14.1"
     val diffx                  = "0.8.2"
     val http4s                 = "0.23.18"
-    val http4sBlaze            = "0.23.13"
+    val http4sBlaze            = "0.23.18"
```
eikek: Rename!

eikek:
Approved by pairing.

PR #1361, opened 2023-02-25
refactor: changes to the handler core functionality

### LINKED ISSUE #1354: TG stops accepting new events after some time
It looks like after a while TG stops accepting new `TRIPLES_GENERATED` events. TG starts responding with `TooManyRequests (429)` which says it's busy processing events it took in the past. It's difficult to say whether TG is in some endless loop that causes the process never finish or an exception that is not handled correctly is raised and simply terminates the thread without clearing the semaphore controlling the number of concurrent processes.

todos:

- [x] take a look at `EventHandlingProcess.maybeReleaseProcess`
- [x] Make BadRequest on event receiving to carry message and be printed on the caller side
- [x] probably to be taken: `EventHandlerWithProcessLimiter.logError`

jachro:
This PR improves significantly the core logic of Event Handlers and make them more robust.


graph-commons/src/main/scala/io/renku/events/consumers/EventHandler.scala
```diff
@@ -51,13 +51,20 @@ abstract class EventHandlerWithProcessLimiter[F[_]: MonadCancelThrow: Logger](
   @annotation.nowarn("cat=unused")
   protected def onPostHandling(event: Event, result: EventSchedulingResult): F[Unit] = ().pure[F]
 
-  final override def tryHandling(request: EventRequestContent): F[EventSchedulingResult] = {
+  final override def tryHandling(request: EventRequestContent): F[EventSchedulingResult] =
+    checkCategory(request)
+      .leftMap(_.pure[F])
+      .as(doCategoryHandling(request))
```
eikek, line 57: Ah good catch! precondition must come after category check...

jachro: Exactly. Debugging to the rescue :)

PR #1364, opened 2023-03-02
fix: TS migration updating renku version pair in the TS

jachro:
This TS migration updates, the so-called, _renku version pair_ in the `migrations` DS in the TS. It was overlooked during the work introducing the V10 migration

### DIFF
--- a/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/V10VersionUpdater.scala
+++ b/triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/V10VersionUpdater.scala
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2023 Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


eikek (approved):

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/V10VersionUpdater.scala
```diff
@@ -0,0 +1,67 @@
...
+private[migrations] object V10VersionUpdater {
+
+  val name: Migration.Name = Migration.Name("V10 version Updater")
+
+  def apply[F[_]: Async: Logger: SparqlQueryTimeRecorder]: F[Migration[F]] = for {
+    implicit0(ru: RenkuUrl)    <- RenkuUrlLoader[F]()
+    compatibility              <- VersionCompatibilityConfig.fromConfigF[F](ConfigFactory.load())
+    migrationsConnectionConfig <- MigrationsConnectionConfig[F]()
+    executionRegister          <- MigrationExecutionRegister[F]
+  } yield new V10VersionUpdater(compatibility.asVersionPair,
```
eikek, line 63: Not sure if I understand: the version pair to update comes from the config - so just in case it stays and later the config changes - would this affect this migration? I.e. could it then update to something else than V10?

jachro: Right, I was thinking about that for a bit. Initially, I thought, let's just set the hardcoded v10 and cli v2.3.0. However, I realised that:
* This new migration will be run only once and straight after the v10 migration is done, which means it will insert the exact versions which were used during the processing.
* In a case when the v10 migration spreads over more than one cli versions (very unlikely multiple schema versions), the migration I added will insert the last versions that were used at the end of the v10 migration. That I think made me to the decision that it's probably better to read the versions from the config rather than hardcode them.
* In a case of a fresh environment (completely new one), this new migration doesn't not really matter as it will insert the same versions as the Re-provisioning migration would have just set.

PR #1366, opened 2023-03-03
fix: v10 migration not to leave projects non-migrated

jachro:
It was found that the v10 migration can leave non-migrated projects due to a mistake in the algorithm finding the projects to do. The fix changes the approach by adding a backlog build step that prepares a list of projects to migrate in the migrations DS. The migration also will be kicked of every single time there are projects needed migration.


triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/v10migration/MigrationNeedChecker.scala
```diff
@@ -0,0 +1,71 @@
...
+  private lazy val query = SparqlQuery.ofUnsafe(
+    show"${MigrationToV10.name} - check migration needed",
+    Prefixes of (schema -> "schema", renku -> "renku", xsd -> "xsd"),
+    s"""|SELECT DISTINCT ?path
+        |WHERE {
+        |  GRAPH ?id {
+        |    ?id a schema:Project;
+        |        schema:schemaVersion '9';
+        |        renku:projectPath ?path.
+        |  }
```
eikek, line 59: Do we also need to check for those without a version property?

eikek (approved):

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/tsmigrationrequest/migrations/v10migration/MigrationNeedChecker.scala
jachro: Actually, here we don't have to or even cannot. The reason is that if we'd do, we'd always say the v10 migration is required.

PR #1376, opened 2023-03-09
feat: DatasetViewed event to be sent using tg api

jachro:
This PR makes Dataset Viewing get noted in the TS.

### DIFF
--- a/triples-generator-api/src/main/scala/io/renku/triplesgenerator/api/events/DatasetViewedEvent.scala
+++ b/triples-generator-api/src/main/scala/io/renku/triplesgenerator/api/events/DatasetViewedEvent.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2023 Swiss Data Science Center (SDSC)
+ * A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+ * Eidgenössische Technische Hochschule Zürich (ETHZ).
+ *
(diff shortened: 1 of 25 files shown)


triples-generator-api/src/main/scala/io/renku/triplesgenerator/api/events/DatasetViewedEvent.scala
```diff
@@ -0,0 +1,69 @@
...
+import io.renku.events.CategoryName
+import io.renku.graph.model.datasets
+
+import java.time.Instant
+
+final case class DatasetViewedEvent(identifier: datasets.Identifier, dateViewed: datasets.DateViewed)
+
+object DatasetViewedEvent {
+
+  def forDataset(identifier: datasets.Identifier, now: () => Instant = () => Instant.now): DatasetViewedEvent =
```
eikek, line 35: cats effect also has this `Clock` typeclass that provides the current time; a bit less convenient but more "pure" 😄

eikek (approved):
Nice! 👍🏼

triples-generator-api/src/main/scala/io/renku/triplesgenerator/api/events/DatasetViewedEvent.scala
jachro: That's true. I somehow forgot about it. Maybe next time :)

PR #1384, opened 2023-03-15
fix: PROJECT_VIEWED to be persisted if contains younger date

jachro:
This PR changes the `PROJECT_VIEWED` event handling so the date carried in the event is persisted only if newer than what's in the TS already.

### DIFF
--- a/entities-viewings-collector/src/main/scala/io/renku/entities/viewings/collector/projects/EventPersister.scala
+++ b/entities-viewings-collector/src/main/scala/io/renku/entities/viewings/collector/projects/EventPersister.scala
@@ -57,7 +57,7 @@ private[viewings] class EventPersisterImpl[F[_]: MonadThrow](tsClient: TSClient[
   private def persistIfOlderOrNone(event: ProjectViewedEvent, projectId: projects.ResourceId) =
     findStoredDate(projectId) >>= {
       case None => insert(projectId, event)
-      case Some(date) if (date compareTo event.dateViewed) < 0 =>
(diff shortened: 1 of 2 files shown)


entities-viewings-collector/src/main/scala/io/renku/entities/viewings/collector/projects/EventPersister.scala
```diff
@@ -51,7 +51,15 @@ private[viewings] class EventPersisterImpl[F[_]: MonadThrow](tsClient: TSClient[
...
     findProjectId(event) >>= {
       case None            => ().pure[F]
-      case Some(projectId) => deleteOldViewedDate(projectId) >> insert(projectId, event)
+      case Some(projectId) => persistIfOlderOrNone(event, projectId)
+    }
+
+  private def persistIfOlderOrNone(event: ProjectViewedEvent, projectId: projects.ResourceId) =
+    findStoredDate(projectId) >>= {
+      case None => insert(projectId, event)
+      case Some(date) if (date compareTo event.dateViewed) < 0 =>
```
eikek: I think using infix is supported for comparables, like `date < event.dateViewed` (not important ofc, just maybe reads a bit easier :))

eikek (approved):

entities-viewings-collector/src/main/scala/io/renku/entities/viewings/collector/projects/EventPersister.scala
jachro: 👍🏻 Changed

PR #1395, opened 2023-03-21
feat: Dataset Details API to accept sameAs or identifier

### LINKED ISSUE #1382: Dataset Details API to accept DS resource id
As a renku-graph user, I'd like the Dataset Details API to accept the whole Dataset ResourceId passed in the URL.

Acceptance criteria:

- [ ] the endpoint not to lose the current ability to work with DS identifier in the URL
- [ ] make the Authorisation work in the case the DS Resource Id is passed
- [ ] the endpoint to work with an Internal Dataset Resource Id passed in the URL
  - [ ] the endpoint to find the parent Project where the given Internal DS exists; in case the DS is modified on the parent Project, some other child Project should be used
- [ ] the endpoint to work with an External Dataset Resource Id passed in the URL
  - [ ] the endpoint to choose an example Dataset and Project where the given External DS exists; it should not be a case where this DS is modified
- [ ] in the case of a Resource Id being passed in the URL, the Resource Id needs to be encoded in a way it's distinguishable from the DS identifier (smth like `int-base64enc(sameAs)`, `ext-base64enc(sameAs)`)

jachro:
closes #1382

### DIFF
--- a/graph-commons/src/test/scala/io/renku/graph/http/server/security/DatasetIdRecordsFinderSpec.scala
+++ b/graph-commons/src/test/scala/io/renku/graph/http/server/security/DatasetIdRecordsFinderSpec.scala
@@ -49,7 +49,7 @@ class DatasetIdRecordsFinderSpec
       )
     }
 
-    "return SecurityRecord with project visibility, path and no member if project is none" in new TestCase {
+    "return SecurityRecord with project visibility, path and no member if project has none" in new TestCase {
 
       val (dataset, project) =
         renkuProjectEntities(anyVisibility)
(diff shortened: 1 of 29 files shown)


graph-commons/src/test/scala/io/renku/graph/http/server/security/DatasetIdRecordsFinderSpec.scala
```diff
@@ -51,10 +49,11 @@ class DatasetIdRecordsFinderSpec
       )
     }
 
-    "return SecurityRecord with project visibility, path and no member is project has none" in new TestCase {
+    "return SecurityRecord with project visibility, path and no member if project is none" in new TestCase {
```
eikek: "... project is none" -> "... project has none"? not sure :) or perhaps "... has no member"

eikek (approved):
Really nice! 👍🏼

PR #1403, opened 2023-03-23
fix: ADD_MIN_PROJECT_INFO to be skipped if project graph exists

### LINKED ISSUE #1400: Flaky Graph Acceptance Tests
It looks like there's some flakiness around getting Project images from the KG API which causes the Graph Acceptance Tests to fail from time to time. So far the problem was observed on: `DatasetsResourcesSpec` and `CommitHistoryChangesSpec`

jachro:
Apparently, it may happen that the `ADD_MIN_PROJECT_INFO` event is delivered after data for a project is put into the TS. In such a case it's both discouraged and unnecessary for the event to be producing data for the TS. This PR adds a check to the `ADD_MIN_PROJECT_INFO` handling process to skip the standard event processing algorithm in case the project exists in the TS already.

closes #1400


eikek (approved):
👍🏼

PR #1404, opened 2023-03-23
feat: sparql string interpolator to resolve iterables considering context

### LINKED ISSUE #1399: Sparql query string interpolator to use proper collection encoding
As a renku graph developer, I'd like the SPARQL query string interpolator to be smart enough to find out what sort of collection encoding needs to be used in a particular context of the part of the query.

Acceptance criteria:

- [x] the interpolator to encode collection items as values wrapped in parenthesis and separated with a '` `' (space) if in the context of the `VALUES` clause
- [x] the interpolator to encode collection items as values separated with a '`,`' (comma) if in the context of the `IN` clause

jachro:
closes #1399

### DIFF
--- a/triples-store-client/src/main/scala/io/renku/triplesstore/client/sparql/StringInterpolator.scala
+++ b/triples-store-client/src/main/scala/io/renku/triplesstore/client/sparql/StringInterpolator.scala
@@ -62,8 +62,9 @@ class StringInterpolator(private val sc: StringContext) {
     ClauseDetector
       .detectClauseContext(sc.parts(idx))
       .map {
-        case ClauseType.VALUES => it.map(makeValue(_, idx)).map(s => s"($s)").mkString(" ")
-        case ClauseType.IN     => it.map(makeValue(_, idx)).mkString(", ")
+        case ClauseType.VALUES_BRACKETED     => it.map(makeValue(_, idx)).map(s => s"($s)").mkString(" ")
+        case ClauseType.VALUES_NOT_BRACKETED => it.map(makeValue(_, idx)).mkString(" ")
+        case ClauseType.IN                   => it.map(makeValue(_, idx)).mkString(", ")
       }
       .getOrElse(sys.error("Iterable cannot be resolved in this context"))
   }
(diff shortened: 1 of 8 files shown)


triples-store-client/src/main/scala/io/renku/triplesstore/client/sparql/StringInterpolator.scala
```diff
@@ -0,0 +1,91 @@
...
+
+    sealed trait ClauseType
+    object ClauseType {
+      case object VALUES extends ClauseType
+      case object IN     extends ClauseType
+    }
+
+    def detectClauseContext(snippet: String): Option[ClauseType] = {
+      val validated = snippet.filter(c => c >= ' ' && c != '|').toUpperCase
+      if (valuesClause matches validated) Some(ClauseType.VALUES)
```
eikek: Wondering what is when there is an `IN` and a `VALUES` in the query where the `IN` comes last. Wouldn't it return in `ClauseType.VALUES`? I thought perhaps it is enough to simply look backwards (maybe via `String#lastIndexOf` and selecting the larger index).

eikek (approved):

triples-store-client/src/main/scala/io/renku/triplesstore/client/sparql/StringInterpolator.scala
jachro: I think we should be fine as the regex is matching at the end of the line. I can maybe write a test verifying that.

eikek: Yes, but isn't there a word matcher `\\w+` that could interfere here? not sure, regex is always a bit hard to verify by eye for me 😄 

Another note, again not sure: couldn't we make the regex simpler by using one class with all possible characters or is the exact sequence important? like `[\\s\\(\\?\\)]*` The single character classes `[\\s]` look unnecessary - why not only `\\s`?

eikek: Ah I think I see now, it wants to match the exact phrase until the open curly. Forget my comment above.

PR #1422, opened 2023-04-03
fix: Throttler not to die losing semaphore permit

jachro:
This is a fix for a really nasty bug we couldn't find for quite some time. It was caused by the wrong usage of a semaphore inside the HTTP client Throttler.


eikek (approved):
🎉 💯

PR #1425, opened 2023-04-06
fix: Retain gitlab images when converting from cli payload

eikek:
The images must be taken from gitlab's project info as well when converting a cli payload.

/deploy #persist


jachro (approved):
Cool! You found it!

PR #1430, opened 2023-04-19
feat: recently viewed projects

### LINKED ISSUE #1389: An API for the recently viewed Projects & Datasets
- create a new endpoint `/recent-entities`
  - the user could be given by the authorization token, then do `/current-user/recent-entities`; or the preferred way is to get the gitlab id (as done with te projects endpoint) and then it would be `/users/:id/recent-entities` 
- use the same datastructures for the response as the cross entity search (`Project` and `Dataset`)
- return the `N` most recent projects/datasets, no other criteria needed
- entrypoint is the `ProjectsViewings` graph
  - it contains a `renku:ProjectViewings` node with edges to nodes with project ids and dataset ids, associating a date and a person to it
  - a potential perf. improvement can be to limit early the number of results in the `ProjectsViewings` graph before resolving the data using the other graphs. but it would return too less results in cases the resolving don't yield a match
  - projects: get the project id and look into the corresponding graph to obtain details
  - datasets: resolve into the `schema:Datasets` graph and use the `link` relation to resolve back to the `DiscoverableDataset`

### Todo:
- [x] create an `openapi.yml` with the endpoint spec to be shared with UI as soon as possible. Write it in scala and let the openapi file be generated
  - optionally have a simple stub to return the advertised structure
- [x] backend for returning recently viewed
  - [x] projects and
  - [x] datasets

eikek:
Implements returning recently viewed entities.

/deploy #persist


entities-viewings-collector/src/main/scala/io/renku/entities/viewings/search/DatasetQuery.scala
```diff
@@ -0,0 +1,129 @@
...
+               |    }
+               |
+               |    {
+               |      SELECT DISTINCT
+               |        ${v.datasetSameAs}
+               |        (GROUP_CONCAT(DISTINCT ?creatorName; separator=',') AS ${v.creatorNames})
+               |        (GROUP_CONCAT(DISTINCT ?keyword; separator=',') AS ${v.keywords})
+               |        (GROUP_CONCAT(DISTINCT ?encodedImageUrl; separator=',') AS ${v.images})
+               |      WHERE {
+               |        Graph schema:Dataset {
```
jachro: Should we maybe use the `GraphClass.Datasets.id` as graph id here (and in few other places in the query)?

entities-viewings-collector/src/test/scala/io/renku/entities/viewings/search/DatasetQuerySpec.scala
```diff
@@ -0,0 +1,137 @@
...
+import io.renku.http.server.security.model.AuthUser
+
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+
+class DatasetQuerySpec extends SearchTestBase {
+
+  it should "return multiple datasets" in {
+    val project1 = renkuProjectEntities(visibilityPublic)
+      .withActivities(activityEntities(stepPlanEntities()))
```
jachro: I'm wondering if we really need to generate activities in these specs? If not, it's always fewer lines of code plus a tiny perf improvement during execution.

entities-viewings-collector/src/test/scala/io/renku/entities/viewings/search/DatasetQuerySpec.scala
```diff
@@ -0,0 +1,137 @@
...
+      .generateOne
+
+    val person1 = personGen.generateOne
+    val person2 = personGen.generateOne
+    upload(projectsDataset, person1, person2)
+    provisionTestProjects(project1, project2).unsafeRunSync()
+
+    val dataset1 = project1.to[entities.RenkuProject.WithoutParent].datasets.head
+    val dataset2 = project2.to[entities.RenkuProject.WithoutParent].datasets.head
+    val userId1  = person1.maybeGitLabId.get
```
jachro: It's not a big deal but if we'd mix in `OptionValues` and use `.value` instead of `.get` we can eventually get slightly more informative failures when they happen.

entities-viewings-collector/src/test/scala/io/renku/entities/viewings/search/ProjectQuerySpec.scala
```diff
@@ -0,0 +1,128 @@
...
+import io.renku.generators.Generators.Implicits._
+import io.renku.http.server.security.model.AuthUser
+
+import java.time.Instant
+
+class ProjectQuerySpec extends SearchTestBase {
+
+  it should "return one project entry if viewed multiple times" in {
+    val project = renkuProjectEntities(visibilityPublic)
+      .withActivities(activityEntities(stepPlanEntities()))
```
jachro, line 32: Personally, for the sake of simplicity, I'd maybe not add activities or datasets to these specs. Forget the comment if I missed something and they are needed here.

entities-viewings-collector/src/test/scala/io/renku/entities/viewings/search/SearchTestBase.scala
```diff
@@ -0,0 +1,102 @@
...
+  lazy val eventUploader: EventUploader[IO] =
+    EventUploader[IO](projectsDSConnectionInfo)
+      .unsafeRunSync()
+
+  lazy val eventPersister: EventPersister[IO] =
+    EventPersister[IO](projectsDSConnectionInfo).unsafeRunSync()
+
+  lazy val tsClient = TSClient[IO](projectsDSConnectionInfo)
+
+  def storeProjectViewed(userId: GitLabId, dateViewed: Instant, path: projects.Path): Unit =
```
jachro, line 71: It's only in the latest PR for the viewing dates deduplication where I'm introducing an `EntityViewings` trait for the purpose of provisioning Viewing Events in our specs. We can maybe consider using them here one the other PR is merged.

entities-viewings-collector/src/test/scala/io/renku/entities/viewings/search/SearchTestBase.scala
```diff
@@ -0,0 +1,102 @@
...
+  def storeDatasetViewed(userId: GitLabId, dateViewed: Instant, ident: datasets.Identifier): Unit =
+    eventUploader.upload(DatasetViewedEvent(ident, dateViewed, userId.some)).unsafeRunSync()
+
+  def personGen: Gen[TestPerson] =
+    personEntities(maybeGitLabIds = personGitLabIds.map(Some(_)))
+
+  def printSparqlQuery(q: SparqlQuery) =
+    println(q.toString.split('\n').zipWithIndex.map(t => f"${t._2}%2d ${t._1}").mkString("\n"))
+
+  def projectDecoder: Decoder[List[SearchEntity.Project]] =
```
jachro, line 83: What do you think about using the `ResultsDecoder` in these decoders? We'd have the `results.bindings` path hardcoded in one place, then.

knowledge-graph/src/main/scala/io/renku/knowledgegraph/MicroserviceRoutes.scala
```diff
@@ -87,9 +90,23 @@ private class MicroserviceRoutes[F[_]: Async](
     (versionRoutes() <+> nonAuthorizedRoutes <+> authorizedRoutes).withMetrics
 
   private lazy val authorizedRoutes: HttpRoutes[F] = authMiddleware {
-    `GET /datasets/*` <+> `GET /entities/*` <+> `GET /projects/*` <+> `GET /users/*`
+    `GET /datasets/*` <+> `GET /entities/*` <+> `GET /projects/*` <+> `GET /users/*` <+> getRecentEntities
   }
 
+  private lazy val getRecentEntities: AuthedRoutes[Option[AuthUser], F] =
```
jachro: I guess we'd need to use the new `MaybeAuthUser` here but the compiler will complain anyway. Actually, is it true that a token is mandatory here? I believe we wouldn't be able to tell much about a current user if the token is not given.

knowledge-graph/src/main/scala/io/renku/knowledgegraph/entities/QueryParamDecoders.scala
```diff
@@ -45,6 +48,37 @@ object QueryParamDecoders {
     val parameterName: String = "type"
   }
 
+  private implicit val recentlyViewedEntityTypesParamDecoder: QueryParamDecoder[RecentEntitiesFinder.EntityType] =
```
jachro, line 50: Do you think it's a good place for the new endpoint params to sit here together with the `entities` endpoint encoders?

triples-generator/src/main/scala/io/renku/triplesgenerator/Microservice.scala
```diff
@@ -64,6 +64,7 @@ object Microservice extends IOMicroservice {
     implicit0(acf: AccessTokenFinder[IO])        <- AccessTokenFinder[IO]()
     implicit0(rp: ReProvisioningStatus[IO])      <- ReProvisioningStatus[IO]()
     config                                       <- parseConfigArgs(args)
+    projectConnConfig                            <- ProjectsConnectionConfig[IO](config)
```
jachro, line 67: I'm wondering if it wouldn't be possible/good to pass the `ProjectsConnectionConfig` in a bit more implicit way. I somehow liked the pureness of these definitions here. The other argument for doing so is that it's rather common for all the instantiations here that they don't need the config for themselves but rather pass them down. What do you think Eike?

entities-viewings-collector/src/main/scala/io/renku/entities/viewings/search/DatasetQuery.scala
eikek: Yes ofc! Forgot these occurrences to replace. Thanks

entities-viewings-collector/src/test/scala/io/renku/entities/viewings/search/DatasetQuerySpec.scala
eikek: right, probably not needed! I'll remove them

knowledge-graph/src/main/scala/io/renku/knowledgegraph/MicroserviceRoutes.scala
eikek: From what I understood here is that the user is required, right. So it probably needs to be `AuthUser` then, thanks!

entities-viewings-collector/src/test/scala/io/renku/entities/viewings/search/ProjectQuerySpec.scala
eikek: I think I can remove the datasets here instead. I thought first, why not add them, just to also see if they are not returned. But can also remove

entities-viewings-collector/src/test/scala/io/renku/entities/viewings/search/SearchTestBase.scala
eikek: Ah yes, thanks. I missed that.

knowledge-graph/src/main/scala/io/renku/knowledgegraph/entities/QueryParamDecoders.scala
eikek: Well, don't have a strong opinion here. I thought that these two could probably share a lot of things, like page/limit and later also entityTypes etc. And they are _used_ outside of that package anyways. But can move them to the sub-package?

triples-generator/src/main/scala/io/renku/triplesgenerator/Microservice.scala
eikek: We can discuss later today if you want. I found it not so nice that every little class is re-reading the config from the filesystem again and again; I tink it terms of "pureness" it is better to pass it along than reading it every time. But the main motivation was to allow different configs for testing purposes. Personally I'd prefer reading the config once and then pass it to where it is needed (or parts of it ofc). Also personal preference :) I don't like using implicits for that, I think it gets easily overused and then very hard to read. If some parts only pass it down without needing actually the data, then perhaps we could introduce the service in need as dependency and not the config for creating it?

knowledge-graph/src/main/scala/io/renku/knowledgegraph/entities/QueryParamDecoders.scala
jachro: Yes, you are right they share a lot. I was just looking at that from the perspective of having things tidy. But, yes, maybe a subpackage that is used by both endpoints would be a good idea.

eikek: I now moved the class one level up. I noticed that it isn't used in that subtree anywhere but solely in the `MicroserviceRoutes` class. Hope that is ok, but can be changed any time obviously.

PR #1450, opened 2023-04-28
fix: project members to be found from project/id/members/all

jachro:
The GL's `/projects/:id/members` API that was used to fetch project members does return only direct project members. The `/projects/:id/members/all` should be used instead.


eikek (approved):
🚀

PR #1465, opened 2023-05-05
feat: Project Status API to work for a non-member caller

### LINKED ISSUE #1462: Project Status API to work for a non-member caller
Currently, the Project Status API fails with 500 when called by a user who is not a member of the project. This is not how it should work.

Acceptance criteria:
- [x] the TR's Associate Project and Token API (TR -> token-repository) should check if the token is valid for the project it is supposed to be associated with (TR to call GL's Project Details API and do nothing in case 401, 403 or 404 or 200 without the `permissions` property is returned); The TR API to return 404 in such a case;
- [x] the Project Status API to call the TR's Associate Project and Token API always if an access token is given
- [x] the Project Status API to always use a token fetched from TR for hook validation
- [x] the Project Status API to return 404 in case hook validation cannot determine if the project hook exists (401, 403, 404 returned from the Project Hook GL API) and EL's Project Events API returns an empty array of events
- [x] the Project Status API to return 200 with `activated=true` in case hook validation cannot determine if the hook exists but EL's Project Events API returns a non-empty array of events
- [x] add renku/graph acceptance tests to verify the logic

jachro:
closes #1462


eikek (approved):
🚀 !!

PR #1508, opened 2023-05-25
refactor: introducing cats-effect-testing-scalatest

jachro:
This PR introduces the `cats-effect-testing-scalatest` lib as an alternative/replacement for the current strategy of testing `IO` code.


eikek (approved):
That is nice!

PR #1518, opened 2023-05-30
fix: PAT leakage for non-activated projects

jachro:
It was identified that we create a new Project Access Token for a project each time the Project Status API is called for a non-activated project. This causes a proliferation of PATs that were unknown for the KG (not stored by token-repository), hence, still active on GL. As a result, the user was able to see a long list of tokens as well as observe reduced responsiveness on certain GL APIs.

This PR fixes the issue by:
1. Moving the PAT revoking functionality to the Delete Project Token API
2. Making sure all Project's PATs are checked while revoking (not just the ones from the first page of the GL's Projects Tokens API response)

### DIFF
--- a/token-repository/src/main/scala/io/renku/tokenrepository/repository/creation/RevokeCandidatesFinder.scala
+++ b/token-repository/src/main/scala/io/renku/tokenrepository/repository/creation/RevokeCandidatesFinder.scala
@@ -23,6 +23,7 @@ import cats.effect.Async
 import cats.syntax.all._
 import io.renku.graph.model.projects
 import io.renku.http.client.{AccessToken, GitLabClient}
+import io.renku.http.rest.paging.model.{Page, PerPage}
 
 import java.time.LocalDate.now
 import java.time.Period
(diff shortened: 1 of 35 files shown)


token-repository/src/main/scala/io/renku/tokenrepository/repository/creation/RevokeCandidatesFinder.scala
```diff
@@ -34,33 +35,59 @@ private trait RevokeCandidatesFinder[F[_]] {
...
+      .iterate(1)(_ + 1)
+      .evalMap(fetch(_, projectId, accessToken))
+      .map { case (tokens, maybeNextPage) =>
+        tokens.filter(renkuTokens).filter(dueToRefresh).map(_._1) -> maybeNextPage
+      }
+      .takeThrough { case (_, maybeNextPage) => maybeNextPage.nonEmpty }
+      .map(_._1)
+      .compile
+      .toList
+      .map(_.flatten)
```
eikek: Since this is only fetched to do something with each token, we could leave it as a stream perhaps? Then we don't need to load everything in memory and can do the action(s) on each chunk within the `Stream` monad.

eikek (approved):
Nice!! 👍🏼

token-repository/src/main/scala/io/renku/tokenrepository/repository/creation/RevokeCandidatesFinder.scala
jachro: Yes, good point. I'll try to follow your suggestion

PR #1519, opened 2023-05-30
feat: Provide name and slug in dataset search response

### LINKED ISSUE #1511: Consistent mapping of dataset slug vs name/title
Historically, datasets only ever had a name which is immutable. Later a user defined "name" was added, called "title". In the schema these are `schema:name` and `renku:slug`. We used to map `slug` to `name`  and `name` to `title`. In issue #1487 changing this mapping was started by introducing a new property `slug` on the dataset that carries the `renku:slug` value. At this point, the cross entity search returns a `name` and a `slug` property carrying the same value (that one of `renku:slug`). The goal is to eventually return the value of `schema:name` as a `name` property and the `renku:slug` as a `slug` property in the api response.

This issue is for applying these changes to all our endpoints, i.e. the dataset details.

Acceptance criteria:
- [x] change the `DatasetSearchInfo` to contain both `name` and `title` (`renku:slug` and `schema:name`)
- [x] add TS migration which simply copies the `schema:name` to all the `DiscoverableDatasets`
- [x] field `name` of the Cross-Entity Search API maps now `schema:name` -> `name` and `renku:slug` -> `slug`. 
  - [x] talk to UI so they can use the new title property
- [x] double check if the `recently-viewed` API follows the same pattern
- [x] align things on the Dataset Details and Project's Datasets API (coordinate with the UI, RP and Omnibenchmark teams)
- [x] **talk to the RP and check if we need to adapt any code on our CLI model decoders** as they could start sending things in a bit different way.
- [x] update README
- [x] remove the `title` property added to the Cross-Entity Search API.

eikek:
Reuses the `name` property on datasets to contain the user defined "title" value  (which is `schema:name` in db) and keeps the `slug` property holding the `renku:slug` value.

Closes: #1511


jachro:
Re-done in #1741

PR #1521, opened 2023-05-31
refactor: Cross-Entity Search's datasets query perf tweaks

jachro:
This PR introduces tiny perf improvements to the Datasets finding query that sits behind the Cross-Entity Search API.


eikek (approved):
Thanks!!

PR #1552, opened 2023-06-13
chore: Recently Viewed API docs fix

### LINKED ISSUE #1551: The "recently-viewed" API route is not documented correctly
### Compare

https://github.com/SwissDataScienceCenter/renku-graph/blob/ff7cda26489de4286542a4a2c5086820b70a6a8b/knowledge-graph/src/main/scala/io/renku/knowledgegraph/MicroserviceRoutes.scala#L98

and

https://github.com/SwissDataScienceCenter/renku-graph/blob/ff7cda26489de4286542a4a2c5086820b70a6a8b/knowledge-graph/src/main/scala/io/renku/knowledgegraph/entities/currentuser/recentlyviewed/EndpointDocs.scala#L46

### Suggested solution

Either the route should be updated to
```
Root / "knowledge-graph" / "entities" / "current-user" / "recently-viewed"
```
or the documentation should be updated to
```
Uri / "current-user" / "recently-viewed"
```

jachro:
closes #1551


eikek (approved):
🙏🏼

PR #1554, opened 2023-06-13
feat: Cache hook validation results

### LINKED ISSUE #1537: Try to make Project Status API cache hook validation checks
Looking at the CPU usage of the token-repository service in response to calls to Project Status API and the hook validation it performs in the background, seems reasonable to cache the hook checks to both improve the response times and lower the mentioned CPU usage.

Acceptance criteria:
- [x] find a relevant cache strategy for the hook validation operations; at first glance, caching the checks by projectId might be a good starting point; in terms of the expiry time maybe something between 1 and 5 seconds could be good.

eikek:
The hook validation result is cached for 10 seconds.

/deploy #persist

closes #1537


jachro (approved):
This looks really cool! Cannot wait to see it in action

PR #1565, opened 2023-06-19
feat: PROJECT_SYNC consumer to send SYNC_REPO_METADATA to TG

### LINKED ISSUE #1443: Sync project metadata without requiring a commit
Currently, some project metadata (like images) are only synced into KG when processing a commit. That means when changing the project image, some "dummy commit" is necessary to ship this data into KG.

Ideally, this metadata should be synced without requiring this dummy commit. There's already a process that periodically syncs some Project data (the `path`) between GL and KG (every 24h as far as I remember) that could be extended for other metadata bits and pieces.

Acceptance criteria:
- [x] `PROJECT_SYNC` event consumer should send a new `SYNC_REPO_METADATA` to TG
  - [x] ensure the `SYNC_REPO_METADATA` is retried on `busy` status received from TG
- [x] update READMEs
- [x] the new event sent to TG to contain the latest event payload
- [x] TG to skip the event if there's no project with the given path in the TS
- [x] TG to extract only relevant Project properties from the payload
- [x] TG to fetch the latest state of the Project data from GL
- [x] TG to update changed values in the TS (for props like `images` the precedence has the payload values over the GL values)
  - Project properties to check are: `visibility`, `images`, `name`, `description` and `keywords`
  - Update should target both the specific `Project` graph as well as the `Projects` graph
 - [x] `DecodingFailure$DecodingFailureImpl DecodingFailure at : Missing required field` for `renku-qa/activation-check`

jachro:
This PR introduces a new `SYNC_REPO_METADATA` event type on TG that syncs Project's name when updated in GL. The new event is issued by the `PROJECT_SYNC` flow to ensure eventual discrepancies are synchronised without the necessity of the user pushing a commit.


generators/src/test/scala/io/renku/generators/Generators.scala
```diff
@@ -395,6 +400,20 @@ object Generators {
...
+
+      def sequence: G[Gen[T]] = {
+
+        @tailrec
+        def genG: G[Gen[T]] = generator.sample match {
+          case Some(value) => value.map(Gen.const)
+          case None        => genG
+        }
+
+        genG
```
eikek: Is there a reason that `sequence` itself is not tail recursive?

triples-generator-api/src/main/scala/io/renku/triplesgenerator/api/events/SyncRepoMetadata.scala
```diff
@@ -0,0 +1,65 @@
...
+  implicit val eventDecoder: Decoder[SyncRepoMetadata] = Decoder.instance { cursor =>
+    import io.renku.tinytypes.json.TinyTypeDecoders._
+
+    lazy val validateCategory = cursor.downField("categoryName").as[CategoryName] >>= {
+      case `categoryName` => ().asRight
+      case other          => DecodingFailure(CustomReason(s"Expected $categoryName but got $other"), cursor).asLeft
+    }
+
+    (validateCategory >> cursor.downField("project").downField("path").as[projects.Path])
+      .map(SyncRepoMetadata(_, maybePayload = None))
```
eikek, line 59: These decoder/encoder never read/write the payload? Just wondering where it is then used?

Ah I think I see now: it is enriched when the event is processed, right? The EL gets the event always without payload and then uses the path to load the latest one.

eikek (approved):

generators/src/test/scala/io/renku/generators/Generators.scala
jachro: Good point. No, not really. I changed this.

triples-generator-api/src/main/scala/io/renku/triplesgenerator/api/events/SyncRepoMetadata.scala
jachro: The thing is that the `payload` is encoded/decoded separately as it's put into a separate part in the multipart request (with an `application/zip` content type).

PR #1577, opened 2023-06-27
feat: avoid concurrent writes to jena

### LINKED ISSUE #1282: Create a service to solve the problem of concurrent TS updates
**Things to consider:**
* Should we process things in a serial way or maybe in parallel on a per-project basis
* Perhaphs we can use Postgres' 13.3.5. Advisory Locks (https://www.postgresql.org/docs/current/explicit-locking.html)
  * https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS
  * https://typelevel.org/cats-effect/docs/std/mutex
  * We'd need to make TG dependent on Postgres
  * We'd need to create TG DB (and user) on Postgres (ConfigMap on the renku chart) and inject the info into TG's config
  * Find out what happens when after acquiring a lock postgres gets restarted. Are the locks released or do we have to somehow release them?

**Acceptance criteria:**
- [ ] do a spike to find out if the concept works (without integrating it to TG)
  - [ ] try to come up with a `Resource` that allows us to do per-projectPath locking
  - [ ] the `Resource` to synchronise through Postgres
  - [ ] do we need to create a table with all the project paths in the new DB?
  - [ ] one we prove the concept works
    - [ ] find out the EventHandlers where the locking needs to be applied
- [ ] should we consider creating an issue to simply the transformation process and remove the deduplication queries?

eikek:
This PR makes the event handlers writing to TS sequentially to avoid concurrent writes that lead to duplicates.

/deploy #persist extra-values=core.sentry.dsn=https://0cc02f3a02d3474d9f7e90915873ab4d@sentry.dev.renku.ch/2,core.sentry.environment=renku-dev,core.sentry.enabled=true

### DIFF
--- a/acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/db/TriplesStore.scala
+++ b/acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/db/TriplesStore.scala
@@ -20,15 +20,12 @@ package io.renku.graph.acceptancetests.db
 
 import cats.effect.{IO, Resource, Temporal}
 import cats.{Applicative, Monad}
-import com.dimafeng.testcontainers.FixedHostPortGenericContainer
 import eu.timepit.refined.auto._
-import io.renku.db.{DBConfigProvider, PostgresContainer}
+import io.renku.db.DBConfigProvider
 import io.renku.triplesgenerator.TgLockDB.SessionResource
 import io.renku.triplesgenerator.{TgLockDB, TgLockDbConfigProvider}
 import io.renku.triplesstore._
-import natchez.Trace.Implicits.noop
 import org.typelevel.log4cats.Logger
-import skunk.Session
 
 import scala.concurrent.duration._
 import scala.util.Try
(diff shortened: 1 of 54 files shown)


acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/db/TriplesStore.scala
```diff
@@ -18,20 +18,56 @@
...
+import skunk.Session
 
 import scala.concurrent.duration._
+import scala.util.Try
 
 object TriplesStore extends InMemoryJena with ProjectsDataset with MigrationsDataset {
 
   protected override val jenaRunMode: JenaRunMode = JenaRunMode.FixedPortContainer(3030)
 
+  private lazy val dbConfig: DBConfigProvider.DBConfig[TgLockDB] =
```
jachro: Wouldn't be better to put the TG's Postgres into a separate object. Or even better, as you suggested on Slack, all EL, TR and the new TG into a single object like `Postgres`? Perhaps, that's a good moment if not requiring too much work.

helm-chart/renku-graph/values.yaml
```diff
@@ -25,6 +25,11 @@ global:
       # Password for the `tokenstorage` user
       # Generate one using: `openssl rand -hex 16`
       value:
+  triplesGenerator:
+    postgresPassword:
+      # Password for the `triplesgenerator` user
+      # Generate one using: `openssl rand -hex 16`
```
jachro: I'm thinking if the `triplesGenerator` section shouldn't be indented two spaces more to be on the same level as `dbEventLog`, I mean under the `graph`. I think, the `tokenRepository` section that is just above the new `triplesGenerator` section may have wrong indentation, too.

helm-chart/renku-graph/values.yaml
```diff
@@ -130,6 +135,7 @@ triplesGenerator:
   ## a demanded number of concurrent triples generation processes
   generationProcessesNumber: 2
   transformationProcessesNumber: 2
+  connectionPool: 2
```
jachro: I'm guessing whether a connection pool of 2 would be good enough for TG. A single connection shouldn't be in use by the lock for long but assuming there are several different events being processed at the same time and some of them might be polling for a lock for a bit, two connections might be tight. But on the other side, maybe it's a good value for now. Just wondering how we'll find out if the setting is fine.

jachro:
This looks really cool! Cannot wait to see it in action :) 

I couldn't approve it as it's still in draft but feel free to merge it whenever you feel it's ready.

helm-chart/renku-graph/values.yaml
eikek: Yes, you are right - I didn't think much when copying these sections, tbh. Perhaps a bit higher value, like 5 or 10? would be better

helm-chart/renku-graph/values.yaml
eikek: Thanks! Love that yaml stuff 😆

acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/flows/TSProvisioning.scala
```diff
@@ -74,7 +75,9 @@ trait TSProvisioning
       sleep((5 seconds).toMillis)
     }
 
-    `wait for events to be processed`(project.id, accessToken, commitIds.size)
+    val items = eventLogClient.getEvents(project.id.asLeft).unsafeRunSync().size
```
jachro: Actually, my idea was slightly different. What I imagined was to pass the expected number of events to the `wait for events to be processed` and only there call the new `eventLogClient.getEvents` and check if the expected number equals the size from the `getEvents`.

eikek: But why would we want to first "guess" and then check? I thought this is more of a helper method to only wait for things so the tests can continue, not really part of the test?

PR #1582, opened 2023-06-29
refactor: dateModified on the Project Details API

### LINKED ISSUE #1564: Add `dateModified` to DS and Project
As a renku graph API user, I'd like the Cross-Entity Search and the Recently Viewed Entities APIs to return info about the modification date.

Acceptance criteria:
- [x] add `dateModified` to all the Project entities in the TS
  - [x] find out if `updated_at` returned by the Single Project GL API gets updated on a commit as well as GL Project metadata changes
  - [x] should we just rely on the mentioned `updated_at` and not even try to compute the date from the payload or perhaps, we'd need to do a max of: project `dateModified` (GL's `updated_at`), datasets' `dateModified`, activities' `startTime` and plans' `dateCreated`
  - [x] make sure relevant flows have the logic of inserting the `dateModified` into the TS:
    - [x] `TRIPLES_GENERATED`
    - [x] `ADD_MIN_PROJECT_INFO`
    - [x] `SYNC_REPO_METADATA`
      - [x] perhaps, we'd need to fetch `dateModified` from TS as an option and create an issue to make it mandatory in the changed flows code later? -> #1575 
  - [x] create a TS migration that adds all the data to TS ~(probably just the Projects graph)~
- [x] make the `GitLabProjectInfo` (project transformation flow) to calculate `dateModified` as max of (`updated_at`, `last_activity_at`, `created_at`)
- [x] go through all the code-base and remove the Option from all the places where Project dateModified is used
  - [x] make sure the `io.renku.graph.acceptancetests.data.Project` don't specify its own `dateModified` but uses the model's `dateModified`
- [x] make the Recently Viewed Entities return the `dateModified` for Project entities
- [x] make the Recently Viewed Entities return the `dateModified` for Dataset entities
- [x] make the Cross-Entity Search return the `dateModified` for Project entities
  - [x] the assumption is that the `date` in the response will still be pointing to `dateCreated` or we should maybe introduce a new `dateCreated` to all the entities and start phasing out the old ambiguous `date` property.
    - [x] if we decide to phase out the `date` we need to:
      - [ ] to talk to the UI and omnibenchmark teams asking to switch to the new `dateCreated`
      - [x] create a new issue to remove the old `date` from the responses
- [x] make the Project Details API return the `dateModified` from the TS (instead the `last_activity_at` from GL)
  - [x] `GitLabProject.dateModified` should go
- [x] make the Cross-Entity Search return the `dateModified` for Dataset entities
- [x] make the Cross-Entity Search to offer sorting by `dateModified`
- [x] after we have the new sortby=dateModified, create an issue to remove the sortBy=date

jachro:
This PR make the Project Details API to return `dateModified` property instead of `updateAt` for consistency reasons.


acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/data/Project.scala
```diff
@@ -27,18 +27,17 @@ import eu.timepit.refined.numeric.Positive
...
 
 import java.net.{MalformedURLException, URL}
-import java.time.Instant
 
 final case class Project(entitiesProject: testentities.RenkuProject,
                          id:              GitLabId,
                          maybeCreator:    Option[ProjectMember],
                          members:         NonEmptyList[ProjectMember],
-                         updatedAt:       DateUpdated,
+                         dateModified:    projects.DateModified,
```
eikek, line 40: just wondering: the descriptions mentions to return `dateCreated` but here it's `dateModified`?

eikek (approved):
👍🏼

acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/data/Project.scala
jachro: Yes, my fault. Description should be saying `dateModified`.

PR #1595, opened 2023-07-06
feat: project.dateModified in search results

### LINKED ISSUE #1564: Add `dateModified` to DS and Project

eikek:
Adopts our code for the fact that `project.dateModified` is mandatory. Adds the new data field to search and recently-viewed responses.

Refs: #1564


entities-search/src/main/scala/io/renku/entities/search/model.scala
```diff
@@ -48,6 +48,7 @@ object model {
         name:             projects.Name,
         visibility:       projects.Visibility,
         date:             projects.DateCreated,
```
jachro, line 50: We could also rename the `date` to `dateCreated` here or leave it for the next PR. Not sure what would work better. Probably it doesn't matter much as the next PR is something we'll do soon anyway. Maybe just worth taking a note.

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/syncrepometadata/processor/model.scala
```diff
@@ -23,6 +23,8 @@ import io.renku.graph.model.images.{Image, ImageUri}
 import io.renku.graph.model.projects
 import io.renku.triplesstore.SparqlQuery
 
+import java.time.Instant
+
 private sealed trait DataExtract {
   val path:              projects.Path
   val name:              projects.Name
```
jachro, line 30: I'm adding a note here but actually, I wanted to put it against the `maybeDateModified` val which is one line below. Should we maybe unwrap it, too, I mean take the Option out of it?

jachro (approved):

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/syncrepometadata/processor/model.scala
eikek: Yes, sure if that is possible? I wasn't sure what is considered "required" in the response from GL.

eikek:
I'm addressing the comments in a next PR, so this can be merged :)

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/syncrepometadata/processor/model.scala
jachro: Ah, yes, you are right. I think we talked about that.

PR #1613, opened 2023-07-19
feat: better 400 reporting for Project Update API

jachro:
This PR improves reporting when the Project Update API responds `400 BAD_REQUEST`. Specifically, in case GL returns 400 response and body like:
```json
{
  "message": {
    "visibility_level": [
      "internal is not allowed in a private group.",
      "internal is not allowed since the fork source project has lower visibility."
    ]
  }
}
```
the Project Update API should return the same amount of details back to the caller.

### DIFF
--- a/graph-commons/src/main/scala/io/renku/data/ErrorMessage.scala
+++ b/graph-commons/src/main/scala/io/renku/data/ErrorMessage.scala
@@ -18,36 +18,59 @@
 
 package io.renku.data
 
+import cats.Show
 import eu.timepit.refined.api.{RefType, Refined}
 import eu.timepit.refined.collection.NonEmpty
+import io.circe.Json
 
 import java.io.{PrintWriter, StringWriter}
 
+sealed trait ErrorMessage {
+  type T
+  val value: T
+  def show: String
+}
+
 object ErrorMessage {
 
-  type ErrorMessage = String Refined NonEmpty
+  final case class StringMessage(value: String Refined NonEmpty) extends ErrorMessage {
+    override type T = String Refined NonEmpty
+    override lazy val show: String = value.value
+    override def equals(obj: Any): Boolean = obj match {
+      case StringMessage(v) => v.value == value.value
+    }
+  }
+  final case class JsonMessage(value: Json) extends ErrorMessage {
+    override type T = Json
+    override lazy val show: String = value.noSpaces
+  }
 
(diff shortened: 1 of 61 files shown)


graph-commons/src/main/scala/io/renku/data/ErrorMessage.scala
```diff
@@ -18,36 +18,59 @@
...
+sealed trait ErrorMessage {
+  type T
+  val value: T
+  def show: String
+}
+
 object ErrorMessage {
 
-  type ErrorMessage = String Refined NonEmpty
+  final case class StringMessage(value: String Refined NonEmpty) extends ErrorMessage {
```
eikek, line 37: Or maybe remove that string message and always use a json message? it could be instantiated by a string to mimic the string message, but would still be json. I think we don't encode these really differently, so if we have a JsonMessage anyways, we could also use it for everything.

graph-commons/src/main/scala/io/renku/data/ErrorMessage.scala
```diff
@@ -18,36 +18,59 @@
...
 import eu.timepit.refined.api.{RefType, Refined}
 import eu.timepit.refined.collection.NonEmpty
+import io.circe.Json
 
 import java.io.{PrintWriter, StringWriter}
 
+sealed trait ErrorMessage {
+  type T
+  val value: T
+  def show: String
```
eikek: I'm wondering if it is necessary to have that string representation - I thought it is only used to create responses?

eikek (approved):
👍🏼

graph-commons/src/main/scala/io/renku/data/ErrorMessage.scala
jachro: We could but sometimes we use `ErrorMessage` factory to create a break and line feed chars free version of String that can be put as a value of tiny types that are stored in Postgres (e.g. `EventMessage.apply`). Sure this logic can be a separate tool, too but it's somehow easier to use/find under an `ErrorMessage` name.
The other usage of the class is to create an error message once and then do both: log it and encode it into the HTTP response.

graph-commons/src/main/scala/io/renku/data/ErrorMessage.scala
jachro: I think I answered your question in the comment above.

PR #1621, opened 2023-07-24
refactor: new Message modelling

jachro:
This PR changes the modelling of the `ErrorMessage` and `InfoMessage`. A new class `Message` with various factory methods has been created. Also JSON encoding of the class has been slightly modified so a new `severity` property was added. For encoding purposes, instead of
`import io.renku.http.InfoMessage._`
`import io.renku.http.ErrorMessage._`
a single 
`import io.renku.data.Message.Codecs._`
should be used.


graph-commons/src/main/scala/io/renku/data/Message.scala
```diff
@@ -0,0 +1,212 @@
...
+    override type T = Json
+    override lazy val show: String = value.noSpaces
+  }
+
+  implicit def show[T <: Message]: Show[T] = Show.show[T] {
+    case m: Message.StringMessage => m.show
+    case m: Message.JsonMessage   => m.show
+  }
+
+  object Codecs {
```
eikek: Why not putting these into the companion objects? Then there wouldn't be a need to add some other import?

eikek (approved):
👍🏼

graph-commons/src/main/scala/io/renku/data/Message.scala
jachro: Done

PR #1642, opened 2023-08-03
feat: Project Create API on KG

### LINKED ISSUE #1635: Add Project Create API on KG
As a renku-graph API user, I'd like the knowledge-graph exposes a Project Create API that performs all the necessary backend operation for a user.

**The problem**
Apparently, the project creation process is slightly complex and requires certain steps to happen in order. While it's possible to do all the work in the UI it's rather a backend task and what's probably more important would need to be replicated in all the renku clients.

**Acceptance criteria:**
- [x] The KG service to expose a new `POST projects` API
  - [x] The API to take all the relevant properties in the payload (to be defined precisely by finding out the bare minimum needed by the Core API)
  - [x] The API to expect a GL auth token can be used against the Core API
  - [x] The API to enforce user authentication in GL
  - [x] The API to block until all the operations are finished
  - [x] The API to return `201 CREATED` for successful creation
  - [x] The API to return `400 BAD_REQUEST` for invalid payload
  - [x] The API to return `400 BAD_REQUEST` in case the project already exists (additional call to GL needs to be done to check this)
  - [x] The API to return `401 UNAUTHORIZED` for auth failures
  - [x] The API to return `500 INTERNAL_SERVER_ERROR` for failures during creation
  - [x] The API to pass eventual error messages coming from the external calls back to the caller
  - [x] ~The API to call the [GL Create Project API](https://docs.gitlab.com/ee/api/projects.html#create-project)~ The project is created in GL by Core's templates.create_project API
  - [x] The API to call the [Core templates.create_project API](https://renku-kg-dev.dev.renku.ch/swagger/#/templates/post_renku_templates_create_project)
  - [x] ~The API to wait for the project to be available (GL's Project Details API to return `200 OK`)~ The API to create the project in the TS - no explicit waiting/polling needed
  - [x] The API to activate the project with the webhook service
  - [x] ~The API to wait for the project to be available in the TS (KG's Project Details API to return `200 OK`)~ No need for waiting for the activation by calling the Project Details API as the project is created in the TS by a direct call
  - [x] ~The API to call the KG's Project Update API to adjust visibility, images and other properties that are not set during creation~
  - [x] Create OpenAPI docs and update the README
  - [x] write ~graph~ renku acceptance tests

jachro:
This PR adds a new Project Create API to the knowledge-graph service.

closes #1635 

/deploy #persist


graph-commons/src/main/scala/io/renku/data/Message.scala
```diff
@@ -66,6 +69,12 @@ object Message extends MessageCodecs {
           .map(toSingleLine)
           .getOrElse(throw new IllegalArgumentException("Message cannot be instantiated with a blank String"))
       }
+
+    def fromJsonUnsafe(message: Json): Message =
+      if (message.isNull || message == Json.obj())
```
eikek: Not sure, an empty array could also occur? If you go further, there could be arrays of empty objects 🤯 😄 But I think we can assume some kind of healthy mind using this, I hope 😄 And if you obtain it from some api, then you'll probably want to see even empty data structures?

knowledge-graph/src/main/scala/io/renku/knowledgegraph/Failure.scala
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.renku.knowledgegraph
+
+import io.renku.data.Message
+import org.http4s.{Response, Status}
+
+private sealed trait Failure extends Exception {
```
eikek, line 24: Seems to me a candidate for the not-yet-existing commons library :) (minus the `toResponse` thing)

knowledge-graph/src/main/scala/io/renku/knowledgegraph/multipart/syntax.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  final implicit class MultipartOps[F[_]: MonadThrow](private val multipart: Multipart[F]) {
+
+    def part(partName: String): F[Part[F]] =
+      findPart(partName).fold(new Exception(s"No '$partName' in the request").raiseError[F, Part[F]])(_.pure[F])
+
+    def findPart(partName: String): Option[Part[F]] =
+      multipart.parts.find(_.name contains partName)
+
+    def findParts(partName: String): List[Part[F]] =
```
eikek: Only for consistency with similar apis, I would expect a `find...` to return an `Option`. What do you think about `filterParts` as a name?

graph-commons/src/main/scala/io/renku/http/client/GitLabClient.scala
```diff
@@ -56,6 +56,10 @@ trait GitLabClient[F[_]] {
       mapResponse: ResponseMappingF[F, ResultType]
   )(implicit maybeAccessToken: Option[AccessToken]): F[ResultType]
 
+  def postMultipart[ResultType](path: Uri, endpointName: String Refined NonEmpty, payload: Multipart[F])(
```
eikek, line 59: I think at some point in the future, it would be nice to have a "real" GitLab-client, not only a generic http clietn with better logging. Now we need to create the correct request at the call site which means every caller must be aware of all the GitLab request details.

eikek (approved):
🚀 🚀

graph-commons/src/main/scala/io/renku/data/Message.scala
jachro: Yes, I think it's good to make it more robust. I added some additional checks.

knowledge-graph/src/main/scala/io/renku/knowledgegraph/Failure.scala
jachro: Yeah!

knowledge-graph/src/main/scala/io/renku/knowledgegraph/multipart/syntax.scala
jachro: Great idea!

graph-commons/src/main/scala/io/renku/http/client/GitLabClient.scala
jachro: I know. I think we were talking about that the other day. I was tempted to maybe do it in the PR but it already grew so big that now I'm happy I didn't do it. I'll create an issue for that.

PR #1652, opened 2023-08-09
feat: renku-core client

### LINKED ISSUE #1631: Project Update API to allow modyfying project keywords
As a renku-graph API user, I'd like the Project Update API to allow modifying keywords of the project of choice.

Acceptance criteria:
- [x] A new Project Update API to be created on the TG that changes the `visibility` and/or `keywords` for the project when specified
  - [x] The API to take the same payload (and have the same constraints) as the KG API (see the next point)
  - [x] The API to update the relevant properties in both the specific `Project` and the `Projects` graphs
- [x] The payload that is accepted by the API needs to be adapted by:
  - [x] The API to take another optional `keywords` property of type `Array[String]` in the payload
    - [x] In the lack of the property, no changes to the project keywords are done
  - [x] The current `visibility` property should become an optional property and in the case it's not provided not changes to the project visibility should be done
  - [x] The API to raise `400 BAD_REQUEST` if neither the `visibility` nor the `keywords` is present in the payload
- [x] The API to call the new TG API to update the TS
- [x] The API not to send the `SYNC_REPO_METADATA` anymore
- [x] The API to change the `keywords` in the project repo by calling the Core's `renku/project.edit` API
  - [x] The flow to update the `keywords` in the project repo is to consist of the following steps:
    - [x] Call the `/api/renku/cache.check_migrations` to fetch `result.core_compatibility_status.project_metadata_version` String value to find out what metadata version the project is; example
    ```
    curl --location 'https://renku-kg-dev.dev.renku.ch/api/renku/cache.migrations_check?git_url=https%3A%2F%2Fgitlab.dev.renku.ch%2Fjakub.chrobasik%2Fanother-test-proj.git' --header 'gitlab-token: <token>'
    ```
    - [x] ~Obtain the `Cookie` header and use it in the subsequent call to the `project.edit` API for perf gains~ maybe in another PR
    - [x] Call the `https://renku-kg-dev.dev.renku.ch/api/renku/versions` to determine Core supports the project version obtained from the previous call. The response should look like this:
     ```json
     {
     "result": {
        "name": "renku-core",
        "versions": [
            {
                "data": {
                    "metadata_version": "10"
                },
                "version": "v2.6.1"
            },
            {
                "data": {
                    "metadata_version": "9"
                },
                "version": "v1.11.4"
            }
        ]
      }
      }
      ```
      - [x] If the project is not supported the KG API to flag that to the caller
      - [ ] ~KG to store in the config the `Core API version` to be used for the next call (probably it will be `1.5` for the time being)~
    - [x] Depending on if that version has the `project.edit` API, call it or fail
      - [x] Probably the call can be done in an async mode by specifying `"is_delayed": true` - actually the call is in sync mode but in a new fiber on KG
      - [x] The payload to include just the `keywords` property (plus all the properties required for the API to work)
- [x] Take care of a case when the default branch is protected. Perhaps we should send some `Conflict` status and not change the TS; or change the TS and say that the data will disappear from TS if the new branch is not merged.
- [x] either make the API to allow changing images in another PR for the issue or create another issue.
- [x] write graph acceptance tests

jachro:
This PR introduces a new module containing a Renku Core client

/deploy #persist


graph-commons/src/test/scala/io/renku/http/client/GitLabClientSpec.scala
```diff
@@ -226,46 +228,40 @@ class GitLabClientSpec
...
         stubFor {
           put(s"/api/v4/$path")
             .withAccessToken(accessToken.some)
-            .withRequestBody(equalTo(s"$propName=$propValue"))
+            .withMultipartRequestBody(
+              aMultipart(partName).withBody(equalTo(partValue))
+            )
+            .withHeader(
+              "Content-Type",
+              containing(s"${multipart.`form-data`.mainType}/${multipart.`form-data`.subType}")
```
eikek: I think you could use ```Renderer.renderString(multipart.`form-data`)``` to render the complete mimetype string. Obviously, just a bean counting detail…

knowledge-graph/README.md
```diff
@@ -811,20 +811,70 @@ Response body example for `Accept: application/ld+json`:
...
+Each of the properties can be either set to a new value or omitted in case there's no new value for it.
+The new values should be sent as a `multipart/form-data` in case there's new image or as JSON in case no update for an image is needed.
+
+The properties that can be updated are:
+* description - possible values are:
+  * `null` for removing the current description
+  * any non-blank String value
+* image - possible values are:
+  * `null` for removing the current image
+  * any file content
```
eikek: again embarassingly bean counting :) should we write "image file" - perhaps there are supported images and some not supported - is there also a max file size?

renku-core-client/src/main/resources/application.conf
```diff
@@ -0,0 +1,10 @@
+
+services {
+
+  renku-core-latest {
+    url = "http://localhost:80"
+    url = ${?RENKU_CORE_LATEST_URL}
+  }
+
+  renku-core-service-urls = ${?RENKU_CORE_SERVICE_URLS}
```
eikek, line 10: There is no sensible default possible, right? Was just thinking how the `RenkuCoreUri#ForSchema` is working - i.e. which kind of value is expected - an array?

renku-core-client/src/main/scala/io/renku/core/client/NestedF.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    override def map[A, B](fa: NestedF[A])(f: A => B): NestedF[B] =
+      Nested {
+        fa.value.map(_.map(f))
+      }
+
+    override def tailRecM[A, B](a: A)(f: A => NestedF[Either[A, B]]): NestedF[B] = Nested {
+      f(a).value >>= {
+        case Result.Success(Right(vb)) => Result.success(vb).pure[F]
+        case Result.Success(Left(va))  => tailRecM(va)(f).value
```
eikek: Not sure if that is really tail recursive? Perhaps we can capture the type class for `Nested` and delegate to it 🤔

renku-core-client/src/main/scala/io/renku/core/client/Result.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    override def flatMap[A, B](fa: Result[A])(f: A => Result[B]): Result[B] =
+      fa match {
+        case Success(a) => f(a)
+        case f: Failure => f
+      }
+
+    override def tailRecM[A, B](a: A)(f: A => Result[Either[A, B]]): Result[B] =
+      flatMap(f(a)) {
+        case Left(v)  => tailRecM[A, B](v)(f)
```
eikek: Same thoughts as for `NestedF` about tail recursiveness... I know we have similar things in the tests, but maybe on prod code we can double check here?

eikek (approved):
Wow!! 👏🏼

renku-core-client/src/main/resources/application.conf
jachro: I see your point. Unfortunately, I'm not sure if we can default to something as we expect the schema version to be encoded into the hostname. I mean, we could default to, let's say `http://renku-core-v10`, but in case you'd like to run it successfully, a relevant entry in the `/etc/hosts` would need to be added. Maybe I'll add such a default.

eikek: Ok, I see, not easy - then it's good to leave it. So an entry in `/etc/hosts` is required? Or "just" a resolvable url? (from the pov of the module)

PR #1661, opened 2023-08-15
feat: Collect auth related data into a separate graph

### LINKED ISSUE #1538: Move authorization data to its own graph
As a renku graph API user, I found KG APIs slow when I provide my access token.

**The problem**
Currently, the APIs authorization goals are achieved by looking into a relevant `Project` graph, `Project` entity's `member` property and matching this data with the caller GitLabId that is found in the relevant `Person` entity in the `Person` graph. This seems to be complicated but, what's by far more important, it's pretty slow. It also doesn't support queries for the owner of a project as it misses this information (the user "role").

**The solution**
Introducing a new graph containing solely the authorization data might be a good solution. Part of the issue would be to come up with an exact shape of the data to be stored in the new graph, however, the main goal would be that it allows fast lookups (probably the amount of data needs to be limited to the essential pieces) while still containing enough info for reasoning about user role on an entity. Looking at the current auth model, info about project, users, access roles might be sufficient.

**Acceptance criteria:**
- [x] create a new module (a bit analog to `entity-search`) as a library to support the new authorization graph
  - [x] A new entity is created for each member of the Project in the Project graph
    ```
    ?memberId a renku:ProjectMember
    ?memberId renku:project projected
    ?memberId renku:person personId
    ?memberId renku:role 'owner'|'maintainer'|'reader'
    ```
    The role can be obtained via the `access_level` property from the response of Project Members All API.
  - [x] Each time in the process, check if the Access Level stored in the TS and fetched from GL is the same and update it in case it's different
- [x] the `MEMBER_SYNC` process is extended to store the new data
- [x] ~~There are also other processes like `ADD_MIN_PROJECT_INFO` and `TRIPLES_GENERATED` that reads the list of project members. Probably these should also add info about the Access Level.~~ We found out, that only MEMBER_SYNC neeeds to do this
  - [x] Quite likely we'd need to add a new `ProjectMember` type which will replace the `Person` type in the `Project.member` property
    - [x] the encoder would need to cope with the new type and encode the new `ProjectMember` entities and keep the `schema:member` unchanged
- [x] ensure the `MEMBER_SYNC`, ~~`ADD_MIN_PROJECT_INFO` and `TRIPLES_GENERATED`~~ store members the `Project` graph (the old way) and to the new auth graph
- [x] ~~add a TS migration that copies over the project members data to the new auth graph~~ We don't do a migration, since one release is used to insert the data over time using the standard machinery
- [x] change the `Authorizer` (specifically the `TSPathRecordsFinder`) to use the data from the new auth graph
- [x] make sure the data is removed from the `project-auth` graph as part of handling the `CleanUpEvent` on TG. Effectively, some necessary additions would need to be done to the `io.renku.triplesgenerator.events.consumers.cleanup.namedgraphs.TSCleaner`.
- [x] find out all the APIs that use the project's members data for authorization purposes and switch them to use data from the new auth graph
- [x] create an issue to remove from the `MEMBER_SYNC`, `ADD_MIN_PROJECT_INFO` and `TRIPLES_GENERATED` to store the members data in the `Project` graphs. Perhaps members can still be stored, but auth related data not.
- [x] remove the project's members data from the `Project` graphs
- [x] do some clean-up in the `MEMBER_SYNC` event handling process

eikek:
This PR introduces another graph in the projects dataset to collect authc/authz related data that comes from GitLab.

/deploy #persist


project-auth/src/main/scala/io/renku/projectauth/ProjectAuthService.scala
```diff
@@ -0,0 +1,147 @@
...
+  private final class Impl[F[_]: MonadThrow](sparqlClient: SparqlClient[F], renkuUrl: RenkuUrl)
+      extends ProjectAuthService[F] {
+    private[this] val graph = Schemas.renku / "ProjectAuth"
+    private implicit val rUrl: RenkuUrl = renkuUrl
+
+    override def remove(slugs: NonEmptyList[Slug]): F[Unit] =
+      sparqlClient.update(sparql"""PREFIX schema: <http://schema.org/>
+                                  |PREFIX renku: <https://swissdatasciencecenter.github.io/renku-ontology#>
+                                  |
+                                  |DELETE { Graph ${graph.asSparql} {?s ?p ?o} }
```
jachro: It's a super tiny thing but I believe you don't need the `.asSparql` on a `Property`. Or from a different side, you we need it, I'd say we should tweak the `sparql` String Interpolator to be able to work with it.

project-auth/src/main/scala/io/renku/projectauth/ProjectAuthService.scala
```diff
@@ -0,0 +1,147 @@
...
+        .map { case (slug, rest) =>
+          val members = rest.toList.flatMap(_._3)
+          val vis     = rest.head.map(_._2)
+          vis.map(v => ProjectAuthData(slug, members.toSet, v))
+        }
+        .unNone
+
+    private def getChunk(limit: Int, offset: Int) =
+      sparqlClient.queryDecode[(Slug, Visibility, Option[ProjectMember])](
+        sparql"""PREFIX schema: <http://schema.org/>
```
jachro, line 119: Do you think it's better to query all the rows and group them on scala side? As you know we can also ask Jena to group them so then you transporting a single row per slug. Just a thought, though :)

project-auth/src/main/scala/io/renku/projectauth/Role.scala
```diff
@@ -0,0 +1,87 @@
...
+ * limitations under the License.
+ */
+
+package io.renku.projectauth
+
+import cats.Order
+import cats.data.NonEmptyList
+import io.circe.{Decoder, Encoder}
+
+sealed trait Role extends Ordered[Role] {
```
jachro: I'm wondering if `Role` shouldn't go to Renku tiny types as a common concept everywhere. I think we already have it implemented in some way in some other place(s). We could maybe reuse the new `Role` and use it everywhere.

project-auth/src/test/scala/io/renku/projectauth/ProjectAuthServiceSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+      for {
+        data <- Generators.projectAuthDataGen.asStream.toIO.take(20).compile.toVector
+        _ <- Stream
+               .emits(data)
+               .through(s.updateAll)
+               .compile
+               .drain
+
+        n <- s.getAll(15).compile.toVector
+        _ = n shouldBe data.sortBy(_.path)
```
jachro: Is that the Project `path`? If so I think it would be good to rename it to `slug`

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/ProjectAuthSync.scala
```diff
@@ -0,0 +1,91 @@
...
+
+    override def syncProject(slug: Slug, members: Set[ProjectMember]): F[Unit] =
+      logger.warn(s"Start looking for visibility for $slug") *>
+        visibilityFinder.find(slug).flatMap {
+          case Some(vis) => syncProject(ProjectAuthData(slug, members, vis))
+          case None      => ().pure[F]
+        }
+
+    override def syncProject(data: ProjectAuthData): F[Unit] =
+      logger.warn(s"Syncing project auth using $data") *>
```
jachro: Why do you want to log it with `warn` severity? Actually, considering that we log on event arrival in the handler, I'd be a bit cautious issuing another log entry here. I can imagine we'll have these two rows printed one after another in the TG logs giving more or less the same info.

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/ProjectAuthSync.scala
```diff
@@ -0,0 +1,91 @@
...
+  private final class Impl[F[_]: Sync](
+      projectAuthService: ProjectAuthService[F],
+      sparqlClient:       ProjectSparqlClient[F]
+  ) extends ProjectAuthSync[F] {
+    implicit val logger: Logger[F] = Slf4jLogger.getLogger[F]
+    private[this] val visibilityFinder: VisibilityFinder[F] =
+      new VisibilityFinder[F](sparqlClient)
+
+    override def syncProject(slug: Slug, members: Set[ProjectMember]): F[Unit] =
+      logger.warn(s"Start looking for visibility for $slug") *>
```
jachro: Keeping in mind we'd log when we sync the data, maybe we can only log on the None case when no project is found.

triples-store-client/src/main/scala/io/renku/triplesstore/client/http/SparqlQuery.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import org.http4s.headers.`Content-Type`
+import org.http4s.{EntityEncoder, MediaType}
+
+trait SparqlQuery {
+  def render: String
+}
+
+object SparqlQuery {
+  final case class Raw(render: String) extends SparqlQuery
```
jachro, line 29: Should it be also possible to new a `SparqlQuery` out of the `Fragment`?

project-auth/src/main/scala/io/renku/projectauth/ProjectAuthService.scala
eikek: I tried both ways and couldn't see a performance gain. Then opted for this approach because there is no string manipulation. But not a strong opinion at all!

project-auth/src/main/scala/io/renku/projectauth/Role.scala
eikek: Ah good point, thanks. I need to look where we already use a role

project-auth/src/test/scala/io/renku/projectauth/ProjectAuthServiceSpec.scala
eikek: Oops, yes! Must be changed

triples-generator/src/main/scala/io/renku/triplesgenerator/events/consumers/ProjectAuthSync.scala
eikek: Oh sorry, forgot to remove this 😅

project-auth/src/main/scala/io/renku/projectauth/ProjectAuthService.scala
jachro: Probably there won't be much of a difference. It's maybe just a matter which CPU, TG's or Jena's would do more cycles :)

jachro (approved):

project-auth/src/main/scala/io/renku/projectauth/Role.scala
jachro: I think there might be something in the Project Details API where we could be reading roles from GL. Not sure, though, how applicable this modelling can be there. If it is not a very good fit, we could leave a separate implementation just there. However, I'd still move the Role to the TinyTypes as I think it becomes something we are going to use more in many places (e.g. the filtering on the API).

PR #1662, opened 2023-08-15
fix: SYNC_REPO_METADATA to use GL project's name not payload's

jachro:
This PR modifies the logic of the `SYNC_REPO_METADATA` process for calculating the Project's name. Before, the name found in the CLI payload had precedence over the GL name. This was causing issues for projects that were forked as in such cases the name in the payload was the name of the parent Project.


eikek (approved):
🙏🏼

PR #1670, opened 2023-08-24
chore: Add query time measurements to sparql client

### LINKED ISSUE #1667: Add time measurement stats to sparqlclient usages
The `SparqlClient` in `triple-store-client` module is very basic and doesn't integrate with the histogram/stats mechanics we have in `graph-commons`. There should be a way to record query times for this approach as well.

eikek:
The newly created sparql client working on the `projects` dataset now can measure execution times and send them to grafana (+ logging).


jachro (approved):
👍🏻

PR #1676, opened 2023-08-28
feat: project auth graph to be provisioned with visibility from GL

jachro:
There was some uncertainty around the source of the project visibility info in the `MEMBER_SYNC` flow that updates the new `project-auth` graph. As it might be that the `MEMBER_SYNC` event is processed before any data for the project exists in the TS, it seems safer not to reach for that to the TS but rather to the source which is GL. This PR modifies the flow to follow this logic.
Additionally, there's an assumption taken that in case no visibility is found in GL, the project auth data in the new graph should be removed.


eikek (approved):
Thanks! 👍🏼

PR #1697, opened 2023-09-06
feat: Integrate project auth graph into cross entity search

### LINKED ISSUE #1538: Move authorization data to its own graph

eikek:
Cross entity search now uses data from `ProjectAuth` graph to filter visible entities.

/deploy #persist

### DIFF
--- a/entities-search/src/main/scala/io/renku/entities/search/DatasetsQuery.scala
+++ b/entities-search/src/main/scala/io/renku/entities/search/DatasetsQuery.scala
@@ -24,7 +24,6 @@ import io.circe.{Decoder, DecodingFailure}
 import io.renku.entities.search.Criteria.Filters
 import io.renku.entities.search.model.{Entity, MatchingScore}
 import io.renku.graph.model._
-import io.renku.graph.model.entities.Person
 import io.renku.graph.model.projects.Visibility
 import io.renku.http.server.security.model.AuthUser
 import io.renku.projectauth.util.SparqlSnippets
@@ -114,7 +113,7 @@ object DatasetsQuery extends EntityQuery[Entity.Dataset] {
           |        ${namespacesPart(criteria.filters.namespaces)}
           |
           |        # access restriction
-          |        ${accessRightsAndVis(criteria.maybeUser, criteria.filters.visibilities)}
+          |        ${accessRightsAndVisibility(criteria.maybeUser, criteria.filters.visibilities)}
           |
           |        # slug and visibility
           |        $slugVisibility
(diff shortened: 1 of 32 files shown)


entities-search/src/main/scala/io/renku/entities/search/DatasetsQuery.scala
```diff
@@ -146,46 +146,11 @@ object DatasetsQuery extends EntityQuery[Entity.Dataset] {
...
-        visibilities match {
-          case v if v.isEmpty || v.contains(projects.Visibility.Public) =>
-            fr"""|?projId renku:projectVisibility ?visibility .
-                 |VALUES (?visibility) { (${projects.Visibility.Public.asObject}) }""".stripMargin
-          case _ =>
-            fr"""|?projId renku:projectVisibility ?visibility .
-                 |VALUES (?visibility) { ('') }""".stripMargin
-        }
-    }
+    sparql"""
```
jachro, line 149: Wow! Such a giant simplification in this query!

jachro (approved):
Cool! 👍🏻

PR #1714, opened 2023-09-19
feat: Munit for the property testing in effectful env

jachro:
This PR introduces MUnit for property-based testing while executed in the cats-effect environment.


eikek (approved):
That look nice!!

PR #1721, opened 2023-09-25
refactor: extracted Histogram.observe to simplify Histogram API

jachro:
This PR:
* adds a new `Histogram.observe` as a common API for labelled and non-labelled Histogram
* removes `Histogram.startTime` as it's perceived as complicated and buggy
* refactors the usage of `MetricsCollector`s


eikek (approved):
👍🏼 👍🏼

PR #1725, opened 2023-09-27
fix: Split query string on underscores

eikek:
The standard tokenizer honors unicode tokenization rules, that won't split on underscores. https://unicode.org/reports/tr29/

/deploy #persist


jachro:
Wow! That looks really promising and such a simple change!

jachro (approved):

eikek:
Thanks for merging :)!

PR #1774, opened 2023-11-03
feat: Secret config reader to work with base64 and ASCII values

### LINKED ISSUE #1768: Make Webhook service and Token Repository more flexible in taking input secret
**Is your feature request related to a problem? Please describe.**
Currently the two services wants as input for the encryption secret (respectively `HOOK_TOKEN_SECRET` and `TOKEN_ENCRYPTION_SECRET`) a base64 encode of an 8/16/32 bytes Hexadecimal number.
As I am currently working toward allowing the Renku Helm chart being able to generate all the secrets needed when not specified in the values file, this requires some ugly workaround with the Helm template language as there is no primitive function that generates random hex numbers.

Also note that the services expects a base64 encoded of the hex, thus when writing the Kubernetes secret, the base hex must be base64 encoded **twice**, which makes the choice very un-intuitive for the Renku admin.

**Describe the solution you'd like**
It would be nice if the two services accepted any string of size [n,m] (for any reasonable integer n,m) of any UTF-8 characters.

**Describe alternatives you've considered**
I have tested the workaround mentioned above here: https://github.com/SwissDataScienceCenter/renku/pull/3327/files#diff-e390929f7a4f02bf329504a332364145c5ddf443ee4a4df274d12565b162cf07 but as I said I would rather avoid it.

**Additional context**
Please and thank you! 😁

jachro:
This PR changes the crypt secret config reader so it can read ASCII and base64 encoded values.

closes #1768 

/deploy

### DIFF
--- a/graph-commons/src/main/scala/io/renku/crypto/AesCrypto.scala
+++ b/graph-commons/src/main/scala/io/renku/crypto/AesCrypto.scala
@@ -26,6 +26,7 @@ import pureconfig.ConfigReader
 import pureconfig.error.FailureReason
 import scodec.bits.ByteVector
 
+import java.nio.charset.CharacterCodingException
 import java.nio.charset.StandardCharsets.UTF_8
 import java.util.Base64
 import javax.crypto.Cipher
@@ -63,7 +64,8 @@ abstract class AesCrypto[F[_]: MonadThrow, NONENCRYPTED, ENCRYPTED](secret: Secr
 object AesCrypto {
 
   final class Secret private (val value: ByteVector) extends AnyVal {
-    override def toString: String = "<sensitive>"
+    override def toString: String                                   = "<sensitive>"
+    def decodeAscii:       Either[CharacterCodingException, String] = value.decodeAscii
   }
 
   object Secret {
--- a/graph-commons/src/main/scala/io/renku/crypto/AesCrypto.scala
+++ b/graph-commons/src/main/scala/io/renku/crypto/AesCrypto.scala
@@ -20,7 +20,7 @@ package io.renku.crypto
 
 import cats.MonadThrow
 import cats.syntax.all._
-import io.renku.config.ConfigLoader.base64ByteVectorReader
+import io.renku.config.ConfigLoader.{asciiByteVectorReader, base64ByteVectorReader}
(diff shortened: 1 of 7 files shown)


graph-commons/src/main/scala/io/renku/crypto/AesCrypto.scala
```diff
@@ -63,31 +64,40 @@ abstract class AesCrypto[F[_]: MonadThrow, NONENCRYPTED, ENCRYPTED](secret: Secr
...
+    private def fromBase64(b64: String): Either[String, Secret] =
       ByteVector.fromBase64Descriptive(b64).flatMap(apply)
 
     def unsafeFromBase64(b64: String): Secret =
       fromBase64(b64).fold(sys.error, identity)
 
     implicit def secretReader: ConfigReader[Secret] =
-      base64ByteVectorReader.emap(bv =>
-        Secret(bv.takeWhile(_ != 10)).leftMap(err =>
+      base64SecretReader orElse asciiSecretReader
```
eikek, line 88: Does this mean, the ascii secret must not be valid base64? This could happen in theory, perhaps?

eikek (approved):
👍🏼

graph-commons/src/main/scala/io/renku/crypto/AesCrypto.scala
jachro: Yes, I know what you mean. It can happen and yes, if it does it's quite unlikely it will be 16, 24 or 32 chars long after decoding. We'll simply fail horrendously. To be honest I don't really see an easy solution for that assuming we'd like to have a single secret entry in the helm config. From the other side, I guess our yat will be the only people who generate the values (in an automatic way), so perhaps that's acceptable. What do you think? I can talk to Alessandro and check if they are happy with that.

eikek: I'm fine of course. Just worth noting, I guess. If it occurs, it will be something hard to find in a year 😄 And might be a support case for deployments not on our control. But agree that it is not very likely. And if we are limited by helm or whatever other tool, I think there is not much we can do.

PR #1783, opened 2023-11-08
feat: Cross-Entity Search results quality improvements

### LINKED ISSUE #1753: Improve the quality of the Cross-Entity Search results
Apparently, the Cross-Entity Search does not work in a way to satisfy our users. For instance searching for:
* `01_primers` does respond with the `01_primers` project, however, querying by `01 primers` returns no results - a small disclaimer; the recent change that was replacing the `StandardTokeninzer` to the `LetterTokenizer` needed to be reverted as was highly problematic for our omnibenchmark friends who use the Search to look for exact keywords containing `_` char.
* `method_omniclustering_pinned` which is an exact keyword of a DS should return this DS, however, as mentioned above, after introducing the `LetterTokenizer` the Search started to return nothing.
* it would be also good to understand why `LetterTokenizer` makes the Search results so unexpected, for instance, it makes DS with the `method_omniclustering_pinned` keyword unfindable.
* it seems reasonable to have a strategy where we tokenise values with the `LetterTokenizer` and additionally add a non-tokenised value to the index as well (obviously if it makes sense to put a non-tokenised value in the first place, e.g. it contains underscores like in `method_omniclustering_pinned`). The value of the `query` parameter would need to go through the same tokeniser, too. The other thing is that probably different tokenisation strategies should be applied to different properties. The mentioned tokeniser should be probably used only for `slug` and `keyword` but the `StandardTokeninzer` to `description`.

Acceptance criteria:
- [ ] Bundle into Jena a new tokenizer that tokenises values with the `LetterTokenizer` and additionally adds a non-tokenised value to the index (the last option from the above)
- [ ] Add a TS migration that updates the Projects DS config in Jena
- [ ] Add a TS migration that reindexes the TS.

I believe there are more cases.

jachro:
This PR:
* defines a custom Lucene Analyzer that divides text at non-letters as well as keeps groups extracted from splitting on whitespaces;
* applies the analyzer to all the indexed properties except from the `schema:description`;
* makes the default query tokenizer to use the same custom analyzer to tokenise user queries;
* adds TS migrations to update the changed `projects.ttl` and kick off reindexing;

closes #1753

### DIFF
--- a/acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/knowledgegraph/CrossEntitiesSearchSpec.scala
+++ b/acceptance-tests/src/test/scala/io/renku/graph/acceptancetests/knowledgegraph/CrossEntitiesSearchSpec.scala
@@ -33,6 +33,7 @@ import io.renku.graph.model.projects.Role
 import io.renku.graph.model.testentities._
 import io.renku.http.client.UrlEncoder._
 import org.http4s.Status.Ok
+import org.scalacheck.Gen.alphaLowerChar
 import tooling.{AcceptanceSpec, ApplicationServices}
 
 class CrossEntitiesSearchSpec extends AcceptanceSpec with ApplicationServices with TSProvisioning {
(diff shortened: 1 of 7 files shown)


eikek (approved):
That looks really nice!

PR #1786, opened 2023-11-10
feat: role filter to replace owned in the Cross-Entity Search API

### LINKED ISSUE #1486: Add owned filter on the Cross-Entity API
Currently, Cross-Entity API allows filtering by `creator`. The API returns only exact matches, however, it's still problematic if there are multiple users having the same name.

Acceptance criteria:
- [x] ~The Cross-Entity API to get a new `owned` filter to work similarly as on the Recently Viewed Entities API~ (see #1488)
- [x] The API to return a `BAD_REQUEST` in case the filter is set but no access token given in the request
- [ ] The Cross-Entity Search API to get a new `role` filter 
  - [ ] the filter to allow multiple values

jachro:
This PR:
* removes the old `owned` filter from the Cross-Entity Search API
* adds a new `role` filter to the Cross-Entity Search API that can take multiple values and allows finding entities where the caller has the specified roles.

closes #1486


eikek (approved):
Nice! 👍🏼

PR #1807, opened 2023-11-29
fix: TG not to write to TS until all exclusive migrations are run

jachro:
This PR:
* introduces a concept of an _exclusive_ TS migration;
* changes the precondition check causing events to be pushed back until all the exclusive migrations are done;
* `DATASET_VIEWED`,  `PROJECT_ACTIVATED` and `PROJECT_VIEWING_DELETION` to be guarded by the same precondition check;
* adds a check preventing `PROJECT_VIEWED` events from being dequeued until all the exclusive migrations are done;

/deploy


eikek (approved):
🙏🏼

PR #1817, opened 2023-12-07
refactor: event-log to use sbt started pg

jachro:
This PR refactors all the EL specs that need Postgres to use SBT started server.


eikek (approved):
🚀 🚀 🚀

PR #1823, opened 2023-12-20
refactor: token-respository-api

jachro:
This PR:
- introduces a new `token-repository-api` module
- removes the `AccessTokenFinder` class and makes all the usages switched to a new `TokenRepositoryClient`


build.sbt
```diff
@@ -231,11 +239,24 @@ lazy val triplesGenerator = project
     AutomateHeaderPlugin
   )
 
+lazy val tokenRepositoryApi = project
+  .in(file("token-repository-api"))
+  .withId("token-repository-api")
+  .settings(commonSettings)
+  .dependsOn(graphCommons % "compile->compile; test->test")
+  .enablePlugins(
+    JavaAppPackaging,
```
eikek: Do we need this on an api module?

eikek (approved):

build.sbt
jachro: Sure, we don't.

PR #1837, opened 2024-01-15
feat: Project Create API to accept templateRef and templateParameters

jachro:
This PR makes the Project Create API:
* to accept a new `projectRef` string part
* to accept a new `projectParameters` JSON part containing an array of key-value pair objects

/deploy


knowledge-graph/src/main/scala/io/renku/knowledgegraph/multipart/syntax.scala
```diff
@@ -126,4 +127,7 @@ object syntax {
           .pure[F]
       }
     }
+
+  implicit def jsonEntityDecoder[F[_]: Concurrent]: EntityDecoder[F, Json] =
```
eikek: We could probably now use `RenkuEntityCodecs`?

eikek (approved):

